Issue #3, opened 2026-07-26
Script to move Gmail inbox emails into chosen label(s) using local Ollama


leonarduk:
## What
Create a script that connects to Gmail, reads messages in the inbox, and moves (labels + archives out of inbox) each message into one or more chosen labels.

## Why
Manually triaging and filing inbox email into labels is repetitive. A script that classifies and files messages automatically would save time and keep the inbox at zero without relying on brittle static Gmail filter rules.

## How
- Authenticate to Gmail via the Gmail API (OAuth2 credentials, read/modify scope).
- List/fetch messages currently in the inbox (optionally filtered by query, e.g. unread or unlabeled).
- For each message, use a **local Ollama model** to decide which label(s) it belongs to, based on the existing set of Gmail labels (subject/sender/snippet/body as input to the prompt).
- Apply the chosen label(s) to the message and remove the `INBOX` label to move it out of the inbox (support assigning to more than one label per message).
- Support a dry-run mode that logs the proposed label assignment without modifying anything.
- Log actions taken (message id/subject → label(s) applied) for auditability.

## Constraints
- Must use a **local Ollama** instance for the classification step — no calls to a hosted/cloud LLM API for this task.
- Must not permanently delete any email; only re-labeling / removing from inbox.
- Should be idempotent — re-running on already-labeled messages should not duplicate work or mis-file them.
- Gmail API credentials/tokens must not be committed to the repo.

## LLM tier
sonnet

## Success looks like
- Running the script against a test/demo Gmail inbox correctly assigns each message to one or more existing labels using the local Ollama model and removes it from the inbox.
- Dry-run mode shows the same classification decisions without mutating any Gmail state.
- Script is documented (usage, required env vars/config, Ollama model name) in the repo README.

## Failure looks like
- Script mislabels or drops messages, or deletes email instead of just relabeling.
- Script silently fails to authenticate or errors out without clear logging.
- Classification calls go to a cloud LLM instead of the local Ollama instance.

PR #7, opened 2026-07-26
Fix Ollama classifier: use JSON-schema response format

leonarduk:
## Summary
Follow-up to #4, found while doing the manual end-to-end verification called out in that PR's test plan.

`ollama_classifier.py` was calling Ollama with `format: "json"`, expecting a JSON *array* back (`["Work"]`). In practice `format: "json"` only guarantees *some* JSON object comes back — against a real local Ollama instance it returned things like `{"Work": true}`, which `_parse_labels` (expecting a list) silently rejected. Every message was being classified as "no label" and left in the inbox — the whole point of the script.

Fixed by using Ollama's structured-output support: pass a JSON schema requiring `{"labels": [...]}` with each item constrained via `enum` to the account's real label names, and parse that shape.

## Verification
Ran `classify_email()` for real against the local Ollama server (not mocked) with both `llama3.2:latest` and `qwen2.5-coder:14b`, using a handful of sample emails and the account's would-be label set (`Work`, `Personal`, `Newsletters`, `Receipts`):
- Before the fix: every email → `[]` (classifier effectively broken)
- After the fix: correctly classified 4/5 with `llama3.2:latest` (small model, one plausible miscategorization) and confirmed the zero-match case returns `[]` as intended
- `python -m pytest tests/ -v` — 23/23 passing (added/updated tests for the new response shape)

Full live run against a real Gmail inbox still needs an OAuth client (`credentials.json`) which isn't set up in this environment, but the classification step — the part this PR touches — is now verified against a real local Ollama server.

Relates to #3


leonarduk:
## Full end-to-end verification against a real Gmail inbox

Ran the OAuth flow, Gmail API calls, and Ollama classification for real (`--dry-run`, so no messages were actually modified):

1. **OAuth**: worked once the app's test-user list included the account being authorized (Google returns `Error 403: access_denied` / "has not completed verification" until you do this — worth calling out to anyone else setting this up).
2. **Gmail API**: correctly listed real inbox messages and the account's actual (large, deeply nested) label tree.
3. **Ollama, first pass** (`llama3.2:latest`, the small 3B default): technically worked (no crashes) but quality was poor — multiple loosely-related labels per email (e.g. a job-application email tagged `Personal/Social/1 - Lucy` + `Finances/Tax/Montpelier`).
4. **Ollama, second pass** (`qwen2.5-coder:14b` + the tightened prompt from the latest commit): single sensible label per email in most cases (job-application emails consistently → `Finances/Job Hunt`), and correctly left ambiguous messages unlabeled (security alert, order confirmation, electoral registration notice) instead of guessing.

Net: the harness (auth → list → classify → label) is solid; classification quality is model/prompt-dependent, which is now called out in the README with a recommendation to use a ≥7B model and always dry-run first against your real label set before a live run.

Did not run a live (non-dry-run) pass — that's a real, hard-to-undo mutation of a real inbox and out of scope for automated verification.

PR #9, opened 2026-07-26
Scope Python CI tests to run per-project, in isolation

leonarduk:
## Summary
Found while investigating adding branch protection: the "Run Tests" job's `pytest projects/ ... || echo "No tests found or tests failed"` swallows its own exit code, so it always reports success regardless of what actually happened.

Removing that swallow exposes that a single repo-wide `pytest projects/` run is genuinely broken:
- Several projects have unqualified `test_server.py`/`server.py` modules with no package structure (no `__init__.py`), so pytest's default import mode throws `import file mismatch` the moment two collide during collection.
- Each project has its own `requirements.txt` (e.g. `mcp>=1.0.0`), but the workflow only ever checked for one at the repo root, so `mcp`-dependent tests fail to even import (`ModuleNotFoundError`).

## Fix
Restructured the `test` job to loop over every `requirements.txt` under `projects/`, install that project's own dependencies, and run `pytest` scoped to just that directory. Skips directories with no test files (avoids pytest's "no tests ran" exit 5 being treated as a failure). The job's real exit code now propagates.

## Verification
Ran the equivalent loop locally (installing each project's deps, running pytest per-directory):
- `01-mcp-server-suite/servers/*`: all pass except `git-mcp-server`'s `test_real_git_lifecycle`, which hit a `UnicodeDecodeError` — **this looked like a Windows console-encoding artifact in my local run**; I could not confirm locally whether it reproduces on `ubuntu-latest` (what this workflow actually runs on). Watching this PR's own CI run will confirm either way.
- `04-gmail-inbox-labeler`: passes (uses the pre-fix test count since #7 hasn't merged yet)
- `web_mcp_server` (no test files): correctly skipped instead of failing

Not fixed here (kept separately scoped): the `lint`/`security` jobs still swallow their exit codes — that's a separate concern from the flaky-on-required-check risk, so left alone as informational-only for now.

## Summary by CodeRabbit

* **Tests**
  * Updated automated testing to run independently for each project.
  * Project-specific dependencies are installed before testing.
  * Projects without test files are skipped automatically.
  * Test results and coverage are reported separately, while failures still affect the overall build status.

### DIFF
--- a/.github/workflows/python-ci.yml
+++ b/.github/workflows/python-ci.yml
@@ -59,14 +59,24 @@ jobs:
 
     - name: Run each project's test suite in isolation
       run: |
-        # Each project directory under projects/ gets its own requirements
-        # install and its own pytest invocation. This avoids two problems
-        # that a single repo-wide `pytest projects/` run hits:
+        # Each project gets its own requirements install and its own pytest
+        # invocation. This avoids two problems that a single repo-wide
+        # `pytest projects/` run hits:
         #   - filename collisions (several projects have their own
         #     test_server.py / server.py with no package structure, which
         #     pytest's default import mode can't disambiguate)
         #   - missing dependencies (a project's own requirements.txt was
         #     never installed when only a root requirements.txt was checked)
+        #
+        # A "project" is identified per test directory, not per
+        # requirements.txt: starting from the directory containing test
+        # files, walk up toward projects/ looking for the nearest
+        # requirements.txt (handles a tests/ subfolder whose requirements.txt
+        # lives in its parent, e.g. projects/04-gmail-inbox-labeler/tests/).
+        # If no requirements.txt is found anywhere in that ancestry, the
+        # project is still tested (just without an install step) instead of
+        # being silently skipped.
+        #
         # git-mcp-server's integration test is excluded: it hardcodes a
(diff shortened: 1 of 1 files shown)


leonarduk:
## Re: DeepSeek review

**§2 (silent skip for projects without requirements.txt)** — valid, fixed in dd2bd29. Switched discovery from "loop over `requirements.txt` files" to "loop over test-file directories, walking up to find the nearest `requirements.txt` (or none)". A project with tests but no `requirements.txt` is now still tested, just without an install step, instead of being dropped. Verified locally with a scratch no-deps project and confirmed the existing `projects/04-gmail-inbox-labeler/tests/` → parent `requirements.txt` resolution still works, then re-ran on real CI (green, see latest "Run Tests" checks on this PR).

**§3 (shared-environment dependency conflicts across sequential `pip install`s)** — agreed this is a real limitation, not addressed in this PR. Filed as a follow-up rather than blocking on it here, since (a) it's not a regression from the pre-fix behavior (that also ran everything in one shared environment, just via a single `pytest projects/` collection that was mostly broken anyway) and (b) actually isolating per-project venvs in CI is a bigger change worth its own PR.

**§5 follow-ups** — noted; will track separately rather than scope-creep this PR further.

Issue #132, opened 2026-08-26
Wire the project into the repo docs and the LinkedIn profile


leonarduk:
## What

Add the LinkedIn Avatar to the root README's project list and repository-structure block, and write
the checklist for putting it on my LinkedIn profile.

## Why

The repo README is the portfolio index — a project that isn't listed there doesn't exist for a
visitor. And the whole milestone exists to produce a link on my profile; that last step deserves to
be written down rather than half-remembered. See design §9.

## How

- Root `README.md`: a `### LinkedIn Avatar` entry in "What I'm Building", matching the existing
  entries' shape (one-paragraph description, status, tech stack), plus `08-linkedin-avatar` in the
  repository-structure block; update the "Last updated" line
- `docs/linkedin-setup.md` in the project: how to add the landing page to the Featured section with a
  custom image and description, how to set the Website field in contact info, and a draft launch post
  explaining what it is and how it was built
- Note in the project README that the twin answers questions about repos in this org, so a new repo
  shows up in its knowledge within a day

## Files Affected

- `README.md`
- `projects/08-linkedin-avatar/README.md`
- `projects/08-linkedin-avatar/docs/linkedin-setup.md`

## Constraints

- Match the tone and structure of the existing root README entries
- No overclaiming — the entry describes what is deployed, not what is planned

## LLM tier

haiku

## Value

Medium Value

## Success looks like

- [ ] The root README lists the project with a working link and accurate status
- [ ] The repository-structure block includes the new directory
- [ ] The LinkedIn setup doc is followable end to end, launch post included

## Failure looks like

- The README describes features that aren't deployed
- The Featured link renders without a preview card because the setup step was skipped

Issue #131, opened 2026-08-26
Add the keep-warm ping workflow


leonarduk:
## What

A scheduled GitHub Action that pings the deployed app every ~14 minutes during waking hours to keep
the free Render instance from sleeping.

## Why

The landing page's prewarm covers a visitor who arrives via the landing page; the ping covers
everyone else and shortens the wake for everyone. See design §9.

## How

- `.github/workflows/avatar-keep-warm.yml`, cron every 14 minutes, restricted to roughly 07:00–23:00
  UTC. Note the corrected arithmetic (design §9): Render gives **750 instance-hours per month
  shared across the workspace**, and a spun-down service consumes none. A 31-day month is 744
  hours, so a 24/7 ping on a single service *would* fit — with ~6 hours of headroom, and only
  if this is the only free service in the workspace. The partial window is a deliberate choice
  to keep that headroom, not an arithmetic necessity
- A single `curl -sSf` at the app URL with a short timeout
- `workflow_dispatch` for manual runs
- Don't fail the workflow on a single failed ping; retry once, then log
- Put the instance-hours arithmetic in a comment, including the fact that widening to 24/7 is
  legitimate if this stays the only free service — re-check the current allowance first

## Files Affected

- `.github/workflows/avatar-keep-warm.yml`

## Constraints

- Must pass `workflow-lint`
- Read-only permissions
- No secrets needed — the URL is public

## LLM tier

haiku

## Value

Low Value — a nice-to-have that meaningfully improves first impressions

## Success looks like

- [ ] The scheduled run succeeds and the app responds without a cold start during the window
- [ ] A transient failure doesn't produce a red X on the repo
- [ ] The schedule stays inside the free instance-hours allowance with headroom to spare

## Failure looks like

- A schedule widened without re-checking the allowance, starving a second free service
- Failing runs producing noise every 14 minutes

Issue #130, opened 2026-08-26
Deploy to Render and document the deployment


leonarduk:
## What

Deploy the app as a Render free web service from the `projects/08-linkedin-avatar` subdirectory, and
write down exactly how, including every environment variable.

## Why

The link is worthless if it isn't live, and a deployment I can't reproduce from notes in six months
is a deployment I'll break. See design §9.

## How

- Render web service: root directory `projects/08-linkedin-avatar`, build
  `pip install -r requirements.txt`, start `python app.py`, free instance
- Environment: `DEEPSEEK_API_KEY`, `AVATAR_PROVIDER` (default `deepseek`), `PUSHOVER_USER`,
  `PUSHOVER_TOKEN`, `GRADIO_SERVER_NAME=0.0.0.0`, `GRADIO_SERVER_PORT=10000`, plus any guardrail
  overrides — see design §10 for the full list. `ANTHROPIC_API_KEY` is only needed if
  `AVATAR_PROVIDER=anthropic` is ever switched on (the documented fallback, not the default path)
- Enable GitHub Pages for the `site/` directory (or document the chosen publishing route)
- Document in `docs/deployment.md`: the settings above, how to rotate a key, how to check logs, what
  the cold-start behaviour is, and how to take the app offline in a hurry
- Record the live URLs in the project README

## Files Affected

- `projects/08-linkedin-avatar/docs/deployment.md`
- `projects/08-linkedin-avatar/README.md`

## Constraints

- No secrets in the repository — every key goes in Render's dashboard
- The doc must be followable by me with no memory of this milestone

## LLM tier

haiku

## Value

High Value

## Success looks like

- [ ] The app is reachable at a public URL and answers a question end to end
- [ ] A contact-capture conversation produces a Pushover notification on my phone
- [ ] The deployment doc lists every variable and the emergency shutdown step
- [ ] A push to `main` triggers a redeploy

## Failure looks like

- A missing environment variable causes a crash loop that only the logs explain
- A key committed to the repo
- No documented way to take it down quickly

Issue #129, opened 2026-08-26
Build the LinkedIn landing page with OG tags and app prewarm (site/index.html)


leonarduk:
## What

A static page — published on GitHub Pages — that LinkedIn actually links to: who this is, what the
twin can do, a "Start chatting" button through to the app, Open Graph tags for the link preview, and
a `fetch()` that wakes the sleeping Render instance on load.

## Why

Render's free tier sleeps after 15 minutes and takes 30–60s to wake. A recruiter who clicks a
LinkedIn link and gets a blank loading page for a minute is gone. A static page loads instantly,
renders a proper preview card on LinkedIn instead of a bare URL, and buys the wake-up time while the
visitor reads. See design §9.

## How

- Single self-contained `site/index.html` — inline CSS, no CDN dependencies
- Open Graph and Twitter card tags: title, description, image, URL
- A 1200x630 preview image (may reuse the repo's `docs/banner.png` styling)
- Above the fold: name, one-line pitch, three example questions, "Start chatting" button
- On load, fire a no-cors `fetch()` at the Render URL to trigger the wake, and show a brief "waking
  up my twin…" state on the button for the first ~20 seconds
- Links to the GitHub repo and the design doc — the page is itself part of the portfolio
- Responsive and legible on mobile; works with JavaScript disabled (button still links through)

## Files Affected

- `projects/08-linkedin-avatar/site/index.html`
- `projects/08-linkedin-avatar/site/og-image.png`

## Constraints

- No third-party scripts, no analytics, no cookies — nothing that needs a consent banner
- Must work as a plain static file, no build step
- The Render URL is a single constant, easy to change

## LLM tier

sonnet

## Value

High Value

## Success looks like

- [ ] Pasting the published URL into a LinkedIn post preview renders title, description and image
- [ ] The page renders correctly at 375px wide
- [ ] With the app asleep, opening the landing page and waiting ~20s means the chat is warm on click
- [ ] The button works with JavaScript disabled

## Failure looks like

- LinkedIn renders a bare URL with no card
- The prewarm fetch throws a visible console error or blocks rendering
- The page needs a build step to publish

PR #133, opened 2026-08-27
docs: add LinkedIn Avatar design and project README

leonarduk:
## What

The design for **LinkedIn Avatar** — an AI twin that talks to visitors about my career and my GitHub
projects, reachable from one link in the Featured section of my LinkedIn profile. Docs only; no
implementation in this PR.

- `projects/08-linkedin-avatar/docs/design.md` — architecture, knowledge pipeline, prompt/token
  budget, tools, safety and cost model, hosting, config table, sequenced build order
- `projects/08-linkedin-avatar/README.md` — project overview

## Why

Starting from the shape of [ed-donner/agents](https://github.com/ed-donner/agents)
`1_foundations/twin` (a Gradio `ChatInterface` over a LinkedIn PDF with two Pushover tools), the
design changes four things and argues for each:

1. **GitHub knowledge** — the differentiator over a CV chatbot. A nightly snapshot of my public repos
   is baked into the cached system prompt, with a `lookup_project` tool for detail on demand. At this
   size a vector store would be architecture theatre.
2. **The LinkedIn PDF never enters this public repo.** It stays gitignored; a builder emits a
   redacted `profile.md` that a human reviews before it is committed, with a `--check` mode CI can
   enforce. A leaked phone number in a public git history is permanent.
3. **Cost and abuse controls are first-class**, not an afterthought — input length cap, per-session
   and per-IP rate limits, and a daily spend kill-switch. This is a public URL with an API key
   behind it.
4. **A static landing page fronts the app** — instant load, a proper Open Graph preview card
   (LinkedIn will not embed an iframe, so the preview *is* the first impression), and a prewarm
   `fetch` that hides Render's 30–60s free-tier cold start.

## How

Model default `claude-haiku-4-5` with prompt caching on the ~20k-token system prompt and
`AVATAR_MODEL` to raise it; a behavioural eval suite decides whether that default holds. Tools are
limited to a push notification and a local read, which bounds the blast radius of a tool jailbreak to
"a visitor can send me a push notification".

## Files Affected

- `projects/08-linkedin-avatar/docs/design.md`
- `projects/08-linkedin-avatar/README.md`

## Related issues

Refs #117, #118, #119, #120, #121, #122, #123, #124, #125, #126, #127, #128, #129, #130, #131, #132

The 16 implementation issues in
[milestone 4](https://github.com/leonarduk/ai-systems-lab/milestone/4) all reference this design doc.
Each is scoped to one or two files with its own tests and labelled `sonnet` (10) or `haiku` (6) —
none needs Opus. Critical path: #117 → #120 → #122 → #124 → #125 → #130.

## Success looks like

- [ ] The design reads as a set of decisions with reasons, not a feature list
- [ ] Someone picking up any issue in milestone 4 can work from the doc without further context

### DIFF
--- a/projects/08-linkedin-avatar/docs/design.md
+++ b/projects/08-linkedin-avatar/docs/design.md
@@ -194,9 +194,11 @@ prompt exceeds it, repo full-records are dropped to index lines, oldest-pushed f
 if it still doesn't fit, that's a build error, not a silent truncation. Token counts come from the
 Anthropic count-tokens endpoint, not a heuristic.
 
-The whole system prompt gets a `cache_control` breakpoint. With a ~15–25k prompt and a 10-turn
-conversation, caching is the difference between "this is a fun demo" and "I've turned it off because
-of the bill".
+The whole system prompt gets a `cache_control` breakpoint, with the **1-hour TTL** rather than the
+default 5-minute one — a visitor reading an answer and typing a follow-up routinely takes longer than
+five minutes, so the short TTL would miss on a large share of turns (§7 has the arithmetic). With a
+~15–25k prompt and a 10-turn conversation, caching is the difference between "this is a fun demo" and
+"I've turned it off because of the bill".
 
 ---
 
@@ -235,7 +237,7 @@ This is a public endpoint with my API key behind it. Four controls, all in `guar
    acceptable weakness for the threat (casual abuse), and adding Redis for this would be the tail
    wagging the dog.
 3. **Daily spend kill-switch** — accumulate `usage` from every response into a running cost estimate;
-   past `AVATAR_DAILY_BUDGET_USD` (default `1.00`) the app stops calling the API and answers with a
+   past `AVATAR_DAILY_BUDGET_USD` (default `5.00`, see §7) the app stops calling the API and answers with a
    fixed "my twin is resting — here's my LinkedIn and email" message. This is the one control that
    must be correct, because it is the only thing standing between a bored visitor and an unbounded
    bill.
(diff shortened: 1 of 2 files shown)


projects/08-linkedin-avatar/docs/design.md
```diff
@@ -0,0 +1,377 @@
...
+about beyond one sentence.
+
+---
+
+## 7. Model and cost
+
+| | |
+|---|---|
+| Provider | Anthropic Messages API (`anthropic` Python SDK) |
+| Default model | `claude-haiku-4-5` — $1 / $5 per MTok in/out, 200K context |
```
leonarduk: **Default model ID may not be valid.** `claude-haiku-4-5` (used here and repeated in the config table at line 331 as the `AVATAR_MODEL` default) looks like it's missing the dated snapshot suffix. Current Haiku 4.5 model IDs are dated (e.g. `claude-haiku-4-5-20251001`), whereas Sonnet/Opus resolve fine as undated aliases — which is presumably why `claude-sonnet-5` on the next line works but this one may not. If the API requires the dated id for Haiku, the first deploy fails with a model-not-found error the moment the avatar tries to answer a question. Worth confirming the exact string against Anthropic's model list before issue #124 bakes it in.

projects/08-linkedin-avatar/docs/design.md
```diff
@@ -0,0 +1,377 @@
...
+| Env override | `AVATAR_MODEL`; `claude-sonnet-5` ($3 / $15) if answer quality needs it |
+| Caching | `cache_control` on the system prompt — cache reads are billed at a fraction of input tokens |
+
+This is a deliberate cost-over-capability choice, not a default: the task is "answer questions from a
+20k-token brief in a consistent voice", which is squarely Haiku-shaped. The eval set (§8) is what
+decides whether that holds; if it doesn't, the fix is one environment variable.
+
+Rough arithmetic for sizing the budget, uncached, at Haiku prices: a 10-turn conversation with a 20k
+system prompt and ~400-token answers is on the order of 200k input + 4k output ≈ **$0.22**. Caching
+the system prompt takes the dominant term down by roughly an order of magnitude. A $1/day cap
```
leonarduk: **$1/day budget seems to undercut its own stated goal.** The doc's own arithmetic just above ($0.22 per uncached 10-turn conversation) means the $1.00/day kill-switch default covers only ~4-5 full conversations before it trips — especially likely early in the day when cache misses are more common. That's a thin margin for the claim a few lines down that "a $1/day cap therefore buys a busy day's worth of genuine visitors." Worth either raising the default, or explicitly caveating that the budget assumes most conversations hit a warm cache.

projects/08-linkedin-avatar/docs/design.md
```diff
@@ -0,0 +1,377 @@
...
+A recruiter who clicks a LinkedIn link and gets a blank loading page for a minute is gone. Two
+mitigations, both in M1:
+
+1. **A static landing page** (`site/index.html`, GitHub Pages) is what LinkedIn actually links to. It
+   loads instantly, carries Open Graph tags so the LinkedIn post/Featured card renders a proper
+   title, description and image instead of a bare URL, explains in two lines what the visitor is
+   about to talk to — and fires a `fetch()` at the Render app the moment it loads, so the instance is
+   waking while the visitor reads. The "Start chatting" button then lands on a warm app.
+2. **A keep-warm ping** — a GitHub Action pinging the app every ~14 minutes during waking hours. One
+   free instance running most of the day fits inside Render's free monthly instance-hours; a 24/7
```
leonarduk: **Keep-warm math looks borderline/possibly backwards.** Render's free tier is commonly ~750 instance-hours/month; a 31-day month is 744 hours. That means a single always-on instance (pinged 24/7) would actually fit within the free allowance on its own — contradicting "a 24/7 ping does not [fit]." (It could still be right if other free services on the account share that pool, but the doc doesn't say that.) Since §7 explicitly warns not to hard-code a remembered pricing number without checking the current page, the same caution seems worth applying here before issue #131 builds a needlessly restricted 'waking hours only' schedule off it.

projects/08-linkedin-avatar/docs/design.md
leonarduk: Checked this against the model docs. `claude-haiku-4-5` is in fact a valid Claude API alias — the docs list `claude-haiku-4-5-20251001` as the ID and `claude-haiku-4-5` as the alias, so the deploy wouldn't have failed with model-not-found. (The nuance behind the instinct: for pre-4.6 models the dateless form is a pointer to a snapshot, whereas for Sonnet 5 / Opus 5 the dateless ID *is* the snapshot.)

But checking turned up a better reason to change it: **Haiku 4.5's retirement commitment is "not sooner than 15 October 2026"** — about seven weeks out. This page is meant to sit unattended on a LinkedIn profile for months, so that's the likeliest way a recruiter ends up looking at an error. Two other corrections fell out at the same time: Sonnet 5's $2/$10 launch pricing is now permanent (the increase to $3/$15 was cancelled), so the doc's "$3/$15" was wrong, and the real gap is only 2× on a workload already capped at a few dollars a day.

So the default is now `claude-sonnet-5`, with Haiku 4.5 kept as a documented cost-down option pinned to the **dated** ID. §7 also notes for #124 that Haiku 4.5 doesn't support `effort` and uses the older `thinking: {type: "enabled", budget_tokens: N}` mode.

Fixed in 1556963; #124 updated so the implementation doesn't bake in the old default.

projects/08-linkedin-avatar/docs/design.md
leonarduk: Agreed — the doc contradicted its own arithmetic, and "buys a busy day's worth of genuine visitors" was an overclaim. Raised the default to **$5.00** and replaced the hand-wave with real figures at verified prices.

Working through it also surfaced a design bug the comment didn't ask about: **the default 5-minute cache TTL is wrong for this workload.** A visitor reading a paragraph of answer and typing a follow-up routinely takes longer than five minutes, so the short TTL would miss on a large share of turns — which is what made the uncached case realistic rather than a worst case. §4 and §7 now specify the 1-hour TTL, which at 2× base input pays for itself after two reads.

Per ten-turn conversation at Sonnet 5 prices, now in §7:

| Scenario | Cost |
|---|---|
| Cached, hit every turn | ≈ $0.17 |
| Every turn misses the cache | ≈ $0.33 |
| No caching at all | ≈ $0.48 |

§7 also now says plainly what the cap is and isn't: a circuit breaker against a scripted abuser, not a capacity plan — if real traffic ever trips it, the answer is to raise it, not to ration. Prices are dated 2026-08-27 in the doc, and #126 now carries them with the same date so the tracker doesn't hard-code a remembered number.

Fixed in 1556963.

projects/08-linkedin-avatar/docs/design.md
leonarduk: You're right and the doc was wrong. Checked Render's free-tier docs: **750 instance-hours per month, shared across the workspace**, and a spun-down service consumes none of them. A 31-day month is 744 hours, so a 24/7 ping on a single service does fit — the "a 24/7 ping does not" claim was simply false.

Two details worth keeping, which is why the window survives in a different form: the allowance is per *workspace*, not per service, so 24/7 only fits while this is the only free service; and it leaves about six hours of headroom for the whole month.

§9 now states the arithmetic and presents the waking-hours window as a deliberate choice to keep that headroom rather than an arithmetic necessity, and says explicitly that widening to 24/7 is legitimate if this stays the only free service. #131 is updated the same way — the "Failure looks like" entry is now "a schedule widened without re-checking the allowance, starving a second free service" rather than the old, wrong "24/7 burns through the free tier by mid-month".

And the same caution the doc applies to pricing applies here: figures checked against the current docs on 2026-08-27, not recalled.

Fixed in 1556963.

PR #135, opened 2026-08-29
fix: repin cicaid-devtools to the wheel that actually exists

### LINKED ISSUE #134: CI broken: cicaid-devtools pinned to nonexistent v0.4.7 release
## What

`requirements-dev.txt` pins `cicaid-devtools` to a wheel URL that returns HTTP 404:

    cicaid-devtools @ https://github.com/leonarduk/cicaid/releases/download/v0.4.7/cicaid_devtools-0.4.7-py3-none-any.whl

`gh release list --repo leonarduk/cicaid` and `git ls-remote --tags` both show the latest published release is `v0.3.0` (2026-08-26) — `v0.4.7` was never tagged or released in that repo.

## Why

Every job that runs `pip install -r requirements-dev.txt` fails at that step in ~6 seconds: Lint Python Code, Run Tests (3.11), Run Tests (3.12), Security Scan, and the DeepSeek/GPT AI review jobs. `main` has been red since commit 7fa90ee (2026-08-22); the last green `python-ci` run was 9dfffd4 (2026-08-15).

## Impact

- No PR gets real lint/test/security signal right now
- Two AI review providers fail before they can post a review
- Contributors following `CONTRIBUTING.md`'s `pip install -r requirements-dev.txt` step hit the same 404 locally

## Fix

Repin to the wheel that actually exists on `v0.3.0`: `cicaid_devtools-0.3.0-py3-none-any.whl`. Verified locally that it installs cleanly and `cicaid --help` runs.

## Success looks like

- [ ] `pip install -r requirements-dev.txt` succeeds
- [ ] `python-ci.yml` is green on `main`

leonarduk:
## What

Repin `cicaid-devtools` in `requirements-dev.txt` from the nonexistent `v0.4.7` wheel to the real `v0.3.0` wheel.

## Why

`v0.4.7` was never tagged or released in `leonarduk/cicaid` — `git ls-remote --tags` and `gh release list` both stop at `v0.3.0`. Every CI job installing dev requirements has 404'd since this was set, taking `main` red since 2026-08-22 (7fa90ee) and blocking Lint, both test-matrix jobs, Security Scan, and the DeepSeek/GPT AI review jobs on every open PR.

## How

Verified locally: `pip install` of the `v0.3.0` wheel succeeds and `cicaid --help` runs.

## Files Affected

- `requirements-dev.txt`

## Success looks like

- [ ] `python-ci.yml` goes green on `main`
- [ ] Open PRs pick up real lint/test/security signal and AI reviews again

Fixes #134

### DIFF
--- a/.github/workflows/_ai-pr-review.yml
+++ b/.github/workflows/_ai-pr-review.yml
@@ -86,8 +86,8 @@ jobs:
 
       - name: Install cicaid-devtools
         env:
-          CICAID_WHEEL_URL: https://github.com/leonarduk/cicaid/releases/download/v0.7.8/cicaid_devtools-0.7.8-py3-none-any.whl
-          CICAID_WHEEL_SHA256: 40b407d76dd94d78f1d961bac73a3f90d57241466a6173392a0a0387ee8bdb27
+          CICAID_WHEEL_URL: https://github.com/leonarduk/cicaid/releases/download/v0.3.0/cicaid_devtools-0.3.0-py3-none-any.whl
+          CICAID_WHEEL_SHA256: 5b113d70851a58a8d6a9bbd6c1160b9569aadac486262140afedc75ecf7f179c
         run: |
           set -euo pipefail
           wheel_file="/tmp/$(basename "$CICAID_WHEEL_URL")"
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
 # Shared repository automation. Keep this pinned to a published cicaid release so
 # local development and CI use the same issue/PR/review commands.
-cicaid-devtools @ https://github.com/leonarduk/cicaid/releases/download/v0.4.7/cicaid_devtools-0.4.7-py3-none-any.whl
+cicaid-devtools @ https://github.com/leonarduk/cicaid/releases/download/v0.3.0/cicaid_devtools-0.3.0-py3-none-any.whl
 
 # Tools used by .github/workflows/python-ci.yml.
 bandit


leonarduk:
Note on the remaining DeepSeek/GPT AI review failures on this PR: they're a separate, deeper bug than the version pin, not something this fix resolves.

**Root cause:** the review-specific modules the workflow imports (`cicaid_devtools.lib.deepseek_review`, `gpt_review`, `review_diff`, `review_comment`, `verdict`, `followups`, etc.) don't exist anywhere in the public `leonarduk/cicaid` package at any version — its own PyPI-style summary says so explicitly: *"GitHub issue/PR plumbing CLI (free shell) - the LLM review/triage engine is cicaid-pro"*. Confirmed by installing v0.3.0 locally: `cicaid_devtools.lib` only ships `github_issue_create`, `github_issues`, `github_repo`, `interactive`, `linked_issue`, `publish_pr` — no review modules.

Those modules live in a separate **private** sibling repo, `leonarduk/cicaid-pro`, which does have a real `v0.7.8` release — that's where the nonexistent-in-public version number in `_ai-pr-review.yml`'s `CICAID_WHEEL_URL` came from. The workflow is just pointed at the wrong repo (`cicaid` instead of `cicaid-pro`).

**Why it can't just be repointed:** `cicaid-pro` is private, so its release assets 404 on an unauthenticated `curl` (confirmed). `_ai-pr-review.yml` currently only has `secrets.GITHUB_TOKEN`, scoped to `ai-systems-lab` alone — it has no read access to `cicaid-pro`.

**To actually fix the AI review jobs:**
1. Add a repo secret (e.g. `CICAID_PRO_TOKEN`) — a PAT with read access to `leonarduk/cicaid-pro`.
2. Update `.github/workflows/_ai-pr-review.yml`'s `CICAID_WHEEL_URL`/`CICAID_WHEEL_SHA256` to point at a real `leonarduk/cicaid-pro` release (latest is `v0.13.1` as of 2026-08-28), and change the `curl` step to send `Authorization: token $CICAID_PRO_TOKEN` (or switch to `gh release download --repo leonarduk/cicaid-pro`, authenticated via that token).

Filing this as a note rather than a fix here since it needs a new secret only repo admin access can add.

PR #138, opened 2026-08-29
fix: install cicaid-devtools-pro from source with an authenticated PAT

### LINKED ISSUE #139: AI review jobs: cicaid-devtools-pro lives in a private repo and needs authenticated install
## What

After #134 fixed the `cicaid-devtools` wheel URL pins, DeepSeek and GPT review still failed. The reason: `review_diff`, `review_comment`, `deepseek_review` and `gpt_review` don't live in the public `cicaid-devtools` package at all — they're in `cicaid-devtools-pro`, in the **private** `leonarduk/cicaid-pro` repo.

## Why

Two separate problems, both real:

1. `_ai-pr-review.yml`'s wheel URL had always pointed at the wrong (public) repo — `cicaid-devtools-pro` was never there to find, at any version.
2. Even pointed at the right repo, `cicaid-pro` stopped attaching wheel release assets partway through its release history, so a wheel-asset pin can't work there regardless of version — it has to be installed from source against a tagged commit.
3. Reaching a private repo at all needs a credential the workflow didn't have: `secrets.gh_token` resolves to the default `GITHUB_TOKEN`, scoped only to `ai-systems-lab`.

## Fix

`CICAID_PRO_TOKEN` (a read-only PAT for `cicaid-pro`) has been added as a repo secret. The install step now does an authenticated `pip install ... @ git+https://github.com/leonarduk/cicaid-pro.git@<tag>`, with the credential injected via a per-command `GIT_CONFIG_COUNT`/`KEY`/`VALUE` override so it never touches the runner's persistent git config or appears in pip's log output.

## Success looks like

- [ ] DeepSeek and GPT review jobs run and post real reviews, not install failures
- [ ] Claude's review job (script-based, no cicaid-devtools-pro dependency) is unaffected

leonarduk:
## What

Install `cicaid-devtools-pro` from source in the AI review workflow, authenticated against the new `CICAID_PRO_TOKEN` repo secret, instead of trying to fetch it as a public wheel asset.

## Why

#134 fixed the `requirements-dev.txt` / `_ai-pr-review.yml` wheel-URL pins so they point at a version of `cicaid-devtools` that actually exists — that's why Lint, both test-matrix legs, Security Scan and CodeQL are green again. But DeepSeek and GPT review kept failing after that fix, for a reason the original issue didn't know about:

- `review_diff`, `review_comment`, `deepseek_review` and `gpt_review` don't live in the public `cicaid-devtools` package at all — they're in `cicaid-devtools-pro`, in the **private** `leonarduk/cicaid-pro` repo. `_ai-pr-review.yml`'s wheel URL had always pointed at the wrong (public) repo.
- Even pointed at the right repo, it wouldn't have worked: `cicaid-pro` stopped attaching wheel release assets from `v0.12.0` onward (2026-08-22 — the same day `main`'s CI went red), so a wheel-asset pin can't work there regardless of version.
- Reaching a private repo at all needs a credential the workflow didn't have: `secrets.gh_token` resolves to the default `GITHUB_TOKEN`, which GitHub scopes only to `ai-systems-lab` and can never read a different repo, private or not.

@leonarduk added `CICAID_PRO_TOKEN` (a read-only PAT for `cicaid-pro`) to unblock this.

## How

- New optional `cicaid_pro_token` secret on the reusable workflow. Optional because Claude's review path uses a vendored `.github/scripts` copy and never needed `cicaid-devtools-pro` — the install step exits cleanly when the token isn't supplied.
- Installs `cicaid-devtools-pro` from source via `pip install "... @ git+https://github.com/leonarduk/cicaid-pro.git@v0.13.1"`, pinned to a tag rather than a wheel asset — robust to `cicaid-pro`'s release pipeline changing again, since it only needs the tag to exist, not a built artifact.
- The token is injected via a global `git config url.insteadOf` rewrite rather than embedded in the pip requirement string, so it never appears in pip's own `Cloning ...` log line on this public repo. The rewrite is removed immediately after in a `trap`.

Verified locally end to end with a personal token standing in for the new secret: the authenticated `git+https` install works, and installing the free `cicaid-devtools` wheel first (already step 1) satisfies `cicaid-devtools-pro`'s dependency on `cicaid-devtools`, which isn't published to PyPI. All four target modules (`review_diff`, `review_comment`, `deepseek_review`, `gpt_review`) import successfully afterward.

## Files Affected

- `.github/workflows/_ai-pr-review.yml`
- `.github/workflows/deepseek-pr-review.yml`
- `.github/workflows/gpt-pr-review.yml`
- `.github/workflows/claude-pr-review.yml`

## Success looks like

- [ ] DeepSeek and GPT review jobs go green on this PR
- [ ] Claude's review job (once `ANTHROPIC_API_KEY` exists) is unaffected — still skips cleanly, doesn't need the new token

Fixes #139

### DIFF
--- a/.github/workflows/_ai-pr-review.yml
+++ b/.github/workflows/_ai-pr-review.yml
@@ -57,6 +57,9 @@ on:
       gh_token:
         description: "Token used for gh CLI calls (PR/issue read and write)"
         required: true
+      cicaid_pro_token:
+        description: "Read-only PAT for the private leonarduk/cicaid-pro repo, used to install cicaid-devtools-pro (the review_module engine). Optional: omit for script-based providers like Claude that don't use a cicaid-devtools review_module."
+        required: false
 
 # github.workflow resolves to the *calling* workflow's name (DeepSeek/GPT PR
 # Review), so this groups per-provider without cross-cancelling other
@@ -94,6 +97,29 @@ jobs:
           curl -fsSL "$CICAID_WHEEL_URL" -o "$wheel_file"
           echo "$CICAID_WHEEL_SHA256  $wheel_file" | sha256sum -c -
           python -m pip install "$wheel_file"
+
+      - name: Install cicaid-devtools-pro
+        env:
+          # cicaid-devtools-pro (the LLM review/triage engine backing the
+          # review_module providers below) lives in the private
+          # leonarduk/cicaid-pro repo and, since v0.12.0, no longer publishes
+          # wheel release assets - it has to be installed from source against
+          # a tagged commit instead. Bump CICAID_PRO_REF when a new pro
+          # release is cut.
+          CICAID_PRO_REF: v0.13.1
+          CICAID_PRO_TOKEN: ${{ secrets.cicaid_pro_token }}
+        run: |
+          set -euo pipefail
+          if [ -z "${CICAID_PRO_TOKEN:-}" ]; then
+            echo "cicaid_pro_token not supplied - skipping cicaid-devtools-pro install (expected for script-based providers, e.g. Claude, that don't use a review_module)."
+            exit 0
+          fi
+          # Route github.com over HTTPS through the token via credential
+          # rewriting rather than embedding it in the pip requirement string,
+          # so it never appears in pip's own "Cloning ..." log line.
+          git config --global "url.https://x-access-token:${CICAID_PRO_TOKEN}@github.com/.insteadOf" "https://github.com/"
+          trap 'git config --global --unset "url.https://x-access-token:${CICAID_PRO_TOKEN}@github.com/.insteadOf"' EXIT
+          python -m pip install "cicaid-devtools-pro @ git+https://github.com/leonarduk/cicaid-pro.git@${CICAID_PRO_REF}"
 
       - name: Get linked issue body
         id: issue
--- a/.github/workflows/_ai-pr-review.yml
+++ b/.github/workflows/_ai-pr-review.yml
@@ -101,11 +101,11 @@ jobs:
       - name: Install cicaid-devtools-pro
         env:
           # cicaid-devtools-pro (the LLM review/triage engine backing the
-          # review_module providers below) lives in the private
-          # leonarduk/cicaid-pro repo and, since v0.12.0, no longer publishes
-          # wheel release assets - it has to be installed from source against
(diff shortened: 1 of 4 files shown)


leonarduk:
Responding to DeepSeek's latest review, since its recommendation this round directly contradicts its own recommendation from two reviews ago — worth resolving explicitly rather than oscillating further.

**Round 2** flagged the original approach (`pip install "... @ git+https://x-access-token:${TOKEN}@github.com/..."`, token embedded in the URL) as a security risk and suggested `actions/checkout` as a safer alternative — quoting: *"clone via `actions/checkout` with the PAT."*

**Round 3** (this one) flags `actions/checkout` — the change made specifically in response to round 2 — and suggests going back to the URL-embedding approach round 2 called a security risk.

Whichever mechanism is used, GitHub Actions has exactly two credential-injection primitives for git over HTTPS: embed it in the URL, or write it to git config (which is what `actions/checkout` does internally — a base64 `Authorization` header under `http.https://github.com/.extraheader`, confirmed in this job's own logs at `git config --local --name-only --get-regexp http\.https\:\/\/github\.com\/\.extraheader`). There is no third option that avoids both.

On the substance of round 3's claim: yes, the token is present in `.git/config` for `cicaid-pro-src` during the job, readable by anything with filesystem access in that job. That's also exactly how this same job's own primary checkout of `ai-systems-lab` has handled `GITHUB_TOKEN` from the start (same `actions/checkout` action, same mechanism, unconditionally, for every job in this file). It's the accepted, standard trust boundary of a single GitHub Actions job — not a gap this PR introduces, and not something either of DeepSeek's two suggested mechanisms actually closes.

Holding `actions/checkout` as the right call here: it reuses well-tested, already-present credential handling (including the post-job cleanup visible in these same logs) instead of a bespoke mechanism, and it's the more idiomatic pattern for exactly this scenario (checking out a private companion repo with a scoped PAT in CI).

Not adopting the other two open suggestions: a pre-flight `git ls-remote` check would duplicate validation `actions/checkout` already performs as its first action (a bad token or ref fails immediately with a clear git error); moving `CICAID_PRO_REF` to a repository variable is inconsistent with this file's existing pinned `CICAID_WHEEL_URL`/`CICAID_WHEEL_SHA256` convention two steps above it.

PR #240, opened 2026-09-16
[Issue #44] Move `GPU_COST_POWER_DEFAULTS` into a separate JSON file for easier user customization.

### LINKED ISSUE #44: Move `GPU_COST_POWER_DEFAULTS` into a separate JSON file for easier user customization.
# Move `GPU_COST_POWER_DEFAULTS` into a separate JSON file for easier user customization

## What

Currently, the GPU power and cost default values (likely a hardcoded dictionary named `GPU_COST_POWER_DEFAULTS` in `llm_cost_comparison.py` or a config module) are defined directly in Python code. This change extracts those defaults into a new JSON file (e.g., `gpu_power_defaults.json`) and updates the loading logic to read from that file. The `pricing.json` file (which holds model pricing) remains untouched; this is a separate configuration concern.

## Why

- **Maintainability**: Users should be able to adjust GPU power consumption and cost assumptions without editing Python code. A dedicated JSON file follows the same pattern already established for `pricing.json`.
- **Correctness risk**: Hardcoded defaults may be overlooked when updating hardware assumptions; externalizing them reduces the chance of stale values.
- **Customization**: Users with non‑standard GPUs or power costs can simply copy and modify the JSON file instead of patching source code.

## How

1. Create `projects/05-llm-cost-comparison/gpu_power_defaults.json` with the same structure as the current `GPU_COST_POWER_DEFAULTS` dictionary. Use the same keys/values and include an `as_of` date and a `note` field for consistency with `pricing.json`.
2. In the module that currently defines `GPU_COST_POWER_DEFAULTS` (likely `llm_cost_comparison.py` or a helper module), replace the hardcoded dict with a function that loads the JSON file using the same pattern already used for `pricing.json` (e.g., `json.load(open(...))` with a sensible default path).
3. Update the `load_pricing` (or equivalent) entry point to also load the GPU defaults file, or load it separately where needed.
4. Update all references that currently use the hardcoded dictionary to use the loaded data (the variable name may change but the API stays the same).
5. Add the new file path to any test fixtures that mock file loading, and ensure tests that verify GPU defaults now read from the JSON file instead of the in‑code constant.
6. Document the new file in the README (e.g., under "Configuration" or "Customising GPU Assumptions").

## Constraints

- **Do not break existing tests**: All 89 tests must still pass. The loading function must return exactly the same data structure (dict with same keys/values) as the hardcoded constant.
- **Keep `pricing.json` unchanged**: This issue is strictly about GPU power/cost defaults. Do not move any pricing data.
- **Backward compatibility**: Existing users who have never modified the code should see no difference in behavior. The new file should be loaded relative to the project root (or the same folder as the script) to match how `pricing.json` is located.
- **Out of scope**: Any changes to the Octopus Agile lookup, timing documentation, or other follow‑ups from the

## Implementation notes

FILES: projects/05-llm-cost-comparison/gpu_power_defaults.json, projects/05-llm-cost-comparison/llm_cost_comparison.py, projects/05-llm-cost-comparison/test_llm_cost_comparison.py, projects/05-llm-cost-comparison/README.md
DONE: `GPU_COST_POWER_DEFAULTS` is extracted into a new `gpu_power_defaults.json` (with `as_of` and `note` fields) loaded via the same pattern as `pricing.json`, all references use the loaded data, existing tests pass unchanged, and the README documents the new file.

leonarduk:
## What
Extracts the hardcoded `GPU_COST_POWER_DEFAULTS` tuple from `llm_cost_comparison.py` into a new `gpu_power_defaults.json` file, and adds a `load_gpu_defaults()` function that reads it. `lookup_gpu_defaults()` now accepts an optional `defaults` argument and loads from the JSON file by default. The README gains a "Customising GPU assumptions" section documenting the file's shape and behavior.

## Why
Users previously had to edit Python source to adjust GPU price/power assumptions for their hardware or region. Externalizing these values into a JSON file mirrors the existing `pricing.json` pattern, reduces the risk of stale hardcoded values, and lets users customize assumptions without touching code.

## Approach
- Followed the established `load_pricing` pattern: a module-level `DEFAULT_GPU_DEFAULTS_PATH` next to the script, `json.load` with a `try/except`, and a tuple return type matching the old constant's shape so downstream code is unchanged.
- Kept a `_FALLBACK_GPU_COST_POWER_DEFAULTS` tuple mirroring the shipped JSON, so a missing or malformed file degrades gracefully and existing users see no behavior change.
- Made `lookup_gpu_defaults` accept an optional `defaults` parameter (defaulting to a fresh load) to keep the function testable without monkeypatching file I/O.
- Skipped malformed individual entries rather than failing the whole load, and fell back to the built-in defaults if the parsed list is empty.
- Rejected keeping the constant as the source of truth with the JSON as an override, since that would leave two places to update and defeat the maintainability goal.

## Testing
Added five tests covering: loading the shipped JSON file, fallback on missing file, fallback on invalid JSON, loading a custom file and matching against it, and passing an explicit defaults tuple to `lookup_gpu_defaults`. Existing tests (including `test_lookup_gpu_defaults_case_insensitive`) continue to pass unchanged.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #44


leonarduk:
Both blocking findings were right, and chasing the first one turned up a related bug the review didn't reach. Addressed in `5d55834`.

**1. `AttributeError` on valid-but-non-object JSON — confirmed and fixed.** `json.load` returns whatever the file holds, and `.get` was called on it unconditionally, so `[1,2,3]`, `"hello"`, `42` and `null` all raised instead of falling back. The top level and the `"gpus"` value are both shape-checked now. Six cases are parametrized in `test_load_gpu_defaults_falls_back_on_wrong_shape`.

**2. "Silent per-entry drop is a real footgun" — agreed, and it was worse than described.** The review called it non-blocking; it is the more damaging of the two bugs. The entire justification for this PR is letting users edit the file, and the old code's response to a user getting an edit slightly wrong was to discard it without a word. If it was their only entry, the shipped defaults came back and the tool behaved exactly as if they had never edited anything. They would see their card fail to match and have no way to find out why.

The rule now distinguishes the two cases, which is the distinction that was missing:

| file state | behaviour |
|---|---|
| absent | fall back, **silently** — this is the out-of-the-box state, not a mistake |
| unparseable / wrong shape / no usable entries | warn on stderr, fall back |
| one entry bad, others fine | warn naming `gpus[<index>]`, keep the rest |

The run is never aborted. That is a deliberate divergence from `load_pricing`, which raises `ConfigError` — and worth stating plainly, because **the PR description's claim to have "followed the established `load_pricing` pattern" was wrong about the part that matters.** `load_pricing` raises on both a missing file and invalid JSON; this silently swallowed both. The divergence is justified, but by the difference in stakes, not by precedent: without pricing nothing can be computed, whereas these values only prefill prompts the user can override, so aborting a costing run over them would be disproportionate.

**3. The `GPU_COST_POWER_DEFAULTS` rename — fixed, though not by aliasing.** Nothing in the repo imports it (grepped), but the review is right that renaming a public-looking symbol is gratuitous here. An alias pointing at `_FALLBACK_GPU_COST_POWER_DEFAULTS` would have been the wrong repair, though: it would hand a caller the hardcoded copy rather than what the user's file says, so anyone who edited the JSON would silently get stale numbers — a quieter bug than the `NameError`. Instead the public name is restored and now holds the *loaded* defaults, so it means what it says and the config file stays the source of truth. `_FALLBACK_...` keeps its underscore, which it has earned.

**4. Drift between the fallback tuple and the shipped JSON — test added.** `test_fallback_gpu_defaults_match_shipped_json` asserts they are equal. Duplication is tolerable only while something enforces the agreement, and nothing did.

**One bug neither review round caught.** Externalising the labels broke the matching the README documents. `lookup_gpu_defaults` upper-cases the detected card name and tests `label in name`, so a lower-case label can never match. That was invisible while the labels lived in Python as an uppercase-by-convention constant; the moment users are invited to type their own, `"rtx 5090"` becomes a completely reasonable thing to write, and it would have silently never matched. Labels are upper-cased at load, and `test_load_gpu_defaults_upper_cases_user_supplied_labels` covers it. `bool` and non-positive numbers are rejected for `cost_usd`/`power_watts` in the same pass — `float(True)` is `1.0`, and a zero-watt GPU is a typo, not a configuration.

**Verification.** All nine new guards were mutation-tested: each was removed in turn and the suite re-run. Every one left a test failing — no survivors. 188 passed (162 on `main`). README updated to describe the warning behaviour, the positive-number rule, and why ordering matters (`RTX 4080 SUPER` before `RTX 4080`, or the shorter label swallows the match).

---

leonarduk:
Right, and a good catch. Fixed in `69f05ca`.

`UnicodeDecodeError` subclasses `ValueError`, not `OSError`, so `except (json.JSONDecodeError, OSError)` missed it entirely and a file saved as UTF-16 or Latin-1 crashed with a traceback. Confirmed the MRO rather than taking it on trust:

```
(<class 'UnicodeDecodeError'>, <class 'UnicodeError'>, <class 'ValueError'>, <class 'Exception'>)
```

The clause now catches `ValueError`, which subsumes `json.JSONDecodeError` as well, with a comment recording why the obvious narrower tuple is wrong. `test_load_gpu_defaults_falls_back_on_non_utf8_file` is parametrized over `utf-16` and `latin-1` and written with `write_bytes`, since `write_text` would only reproduce it by accident.

What makes this finding worth more than the fix is that it is a contract violation I introduced. The README paragraph added in the previous round says a file that is "present but unusable" warns and falls back; a non-UTF-8 file is exactly that, and the code did not honour it. Widening the documented promise while leaving a decode path outside the handler is the same shape of error as the original `.get`-on-a-list — a claim in prose that the `except` clause does not actually cover.

Mutation-verified: restoring `(json.JSONDecodeError, OSError)` fails the new test and nothing else.

190 passed (162 on `main`).

On the follow-up suggestion about `load_pricing`: I would leave its raise-on-error behaviour as it is. The asymmetry is deliberate and is now argued in `load_gpu_defaults`'s own docstring — pricing is load-bearing for every number the tool prints, so continuing without it would produce a comparison of nothing, while these defaults only prefill overridable prompts. Worth a comment near `load_pricing` pointing at that reasoning, but that belongs in its own change rather than widening this one.

---

PR #248, opened 2026-09-16
[Issue #49] Consider validating that `input_per_million` and `output_per_million` in `pricing.json` are positive numbers (currently only presence is checked).

### LINKED ISSUE #49: Consider validating that `input_per_million` and `output_per_million` in `pricing.json` are positive numbers (currently only presence is checked).
## What

Add validation in the pricing-loading logic (the function that reads `pricing.json`) to ensure that the values of `input_per_million` and `output_per_million` are **positive numbers** (strictly greater than 0). Currently the code only checks that the keys exist, not their type or value range.

## Why

* **Correctness**: Negative or zero values would produce nonsensical cost calculations (zero or negative costs) without warning, silently propagating errors downstream.
* **Maintainability**: Explicit validation catches misconfigurations early, making it easier for developers and automated agents to detect malformed pricing data.
* **Agent safety**: An LLM or script that auto‑generates `pricing.json` could inadvertently insert zero/negative values; validation prevents that from going unnoticed.

## How

1. Locate the function that parses `pricing.json` (likely in the main cost‑calculation module, e.g. `cost_calculator.py` or a config loader).  
2. After the JSON is loaded and the `input_per_million` / `output_per_million` keys are confirmed present, add a check that both values are **numeric** (`isinstance(val, (int, float))`) and **greater than 0**.  
3. Raise a clear `ValueError` (or `ConfigError`) with a message like `"input_per_million for model 'X' must be a positive number, got {value}"`.  
4. Optionally, also validate that the values are not `NaN` or infinity (using `math.isfinite()`).

No changes to other data fields, the `as_of` date, or the test suite beyond adding a test case for invalid pricing values (following the existing test pattern).

## Constraints

* Must not break any existing test cases (all current `pricing.json` examples used in tests have valid positive values).
* The validation must run **after** key existence checks so the order of error messages remains logical.
* Do **not** change the schema or structure of `pricing.json` – only add validation logic.
* Out of scope: validating any other pricing fields (e.g. `model`, `provider`) or the `as_of` date.

## LLM tier

**Haiku**. The change is small and mechanical (add a guard clause with a type check and comparison), but it requires reading the surrounding code to place the check correctly and to craft the error message in a consistent style. Haiku can handle that with minimal context; a local‑7b might misplace the check or miss edge cases.

## Success looks like

* A `pricing.json` with `"input_per_million": 0` or `-5` causes the program to exit with an informative error message.
* All existing tests still pass.
* The error message follows the same convention as other config‑validation errors in the codebase.

## Failure looks like

* The validation is added after cost calculations (i.e. the check runs too late to prevent bad arithmetic).
* The check is too restrictive (e.g. rejects `0.0001`, which is a valid positive number).
* Existing tests break because the validation is placed in a location that runs during import or in a path not expected by tests.

## Implementation notes

DONE: The pricing-loading function in `llm_cost_comparison.py` raises a clear `ValueError` when `input_per_million` or `output_per_million` is non-numeric, non-finite, or ≤ 0, and a new test case in `test_llm_cost_comparison.py` covers the invalid-value path while all existing tests continue to pass.

leonarduk:
## What
Added a `_validate_pricing` helper that runs at pricing-load time and rejects any model whose `input_per_million` or `output_per_million` is missing, non-numeric, non-finite (`NaN`/`inf`), or non-positive (`<= 0`). The existing numeric-type check in `build_hosted_rows` was removed since validation now happens earlier, in `load_pricing`.

## Why
Previously only key presence was checked, so a `pricing.json` with `0`, a negative value, or a quoted string like `"5.0"` would silently flow into cost math and produce zero/negative/nonsensical monthly costs with no warning. This is especially risky when pricing files are auto-generated by scripts or LLMs. Validating at load time surfaces misconfigurations early with a clear, field-named error.

## Approach
Validation was placed inside `load_pricing`, immediately after JSON parsing and before the dict is returned, so every consumer benefits and no cost arithmetic can run on bad data. The check uses `isinstance(value, (int, float))` with an explicit `bool` exclusion (since `bool` is a subclass of `int`), plus `math.isfinite` and `value <= 0` to catch `NaN`, `inf`, zero, and negatives. Errors are raised as `ConfigError` to match the existing convention for config problems. The redundant type check in `build_hosted_rows` was removed to avoid duplicated logic and keep a single source of truth. Alternatives considered: validating lazily at each call site (rejected — easy to miss paths and runs too late), and adding a JSON schema (rejected — heavier than needed for two fields).

## Testing
Added parametrized tests covering `0`, `-5`, `0.0`, and `-0.01` for `input_per_million`, a separate case for a non-positive `output_per_million`, a non-numeric (`"5.0"`) value, and a positive-control test confirming a tiny-but-valid rate (`0.0001`) is accepted. All existing tests continue to pass.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #49


leonarduk:
Merged `main` and reworked the change around what landed there in the meantime. `1945b6a`.

**The PR is no longer redundant, but for a reason worth stating precisely.** `main` has since grown `_validate_pricing_model`, which does everything this PR's `_validate_pricing` did — bool, `NaN`/`inf`, non-positive — so at first glance this PR was superseded. It was not. That helper's docstring says:

> Shared by `load_pricing` (via `_validate_pricing`) and `build_hosted_rows` so both paths enforce identical semantics and raise the same `ConfigError` message.

**`_validate_pricing` did not exist on `main`, and `load_pricing` called nothing.** Half that sentence described code that was never written. A zero or negative rate in `pricing.json` was still accepted at load time and only caught later, if it happened to reach `build_hosted_rows` — which the issue's own framing about auto-generated pricing files makes the likely case to miss. This PR supplies the missing half, so the docstring becomes true.

It supplies it as a loop delegating to the shared helper, not as the second independent implementation the PR originally carried. Two copies of the same rule with two different message formats is precisely the drift the helper exists to prevent.

**Kept `build_hosted_rows`'s validation call; the original PR deleted it.** The stated reasoning was "removed since validation now happens earlier" — reasonable when written, wrong now. `main` added that call deliberately, documented as defending a hand-constructed dict that bypassed `load_pricing` entirely, and there is a test calling `build_hosted_rows` directly to prove it. Deleting it would have reopened that hole and turned a green test red.

**One thing neither the PR nor `main` got right: the message.** Both reported a rate of `0` or `-5` as

> pricing model 'claude/opus-5' is missing a numeric input_per_million

The field is present and it is numeric. That message sends the user hunting for an absent key when the real problem is a value they can see. The two failure modes are different mistakes and now read differently — a missing or non-numeric field keeps the original wording, while a present-but-unusable rate gets the PR's better `"input_per_million for model 'claude/opus-5' must be a positive number, got 0"`. The two `build_hosted_rows` tests are reparametrized so each case asserts whichever message it warrants, rather than one blanket string that was wrong for four of the seven.

**Two tests added to hold the line:**

- `test_load_pricing_and_build_hosted_rows_report_the_same_error` runs all seven rejected shapes through both entry points and asserts the messages are byte-identical. The docstring's claim is now enforced rather than believed — which matters, since believing it is what let the gap sit there.
- `test_load_pricing_validates_every_model_not_just_the_first` puts a bad rate behind two good models in a second provider. Validating only the first model would pass every other test in the file and buy nothing for a real multi-provider file.

**Verification.** Six guards mutation-tested — the `load_pricing` call, the `build_hosted_rows` call, bool rejection, `isfinite`, the positive check, and full iteration (truncated to the first model). No survivors. 177 passed (162 on `main`).

---

PR #257, opened 2026-09-16
[Issue #52] In `build_local_row`, when `hours_needed > HOURS_PER_MONTH`, the cost for owned hardware is computed as if a single machine can run more than 24/7, which may underestimate cost; the note warns but the logic could be improved (e.g., require multiple cards or limit to max monthly tokens).

### LINKED ISSUE #52: In `build_local_row`, when `hours_needed > HOURS_PER_MONTH`, the cost for owned hardware is computed as if a single machine can run more than 24/7, which may underestimate cost; the note warns but the logic could be improved (e.g., require multiple cards or limit to max monthly tokens).
## Problem

In `build_local_row` (file location not specified in review, likely in a cost‑estimation module), when `hours_needed > HOURS_PER_MONTH`, the current logic computes the cost for owned hardware as if a single machine can run for the entire `hours_needed` period (i.e., more than 24/7). This under‑estimates the true cost because hardware cannot physically run more than one month’s worth of hours on a single card (assuming `HOURS_PER_MONTH` represents the maximum monthly runtime of one machine). A warning note exists but the logic itself remains incorrect.

## Motivation

- **Correctness risk**: Users who enter workloads requiring multiple months of computation will see an artificially low cost for owning hardware, potentially leading to incorrect purchase decisions.
- **Maintainability**: Relying on a warning note is fragile; future readers may not notice the caveat.
- **Agent confusion**: Automated pricing comparisons (e.g., in CI or interactive mode) could produce misleading results.

## Proposed Solution

Modify `build_local_row` to cap the effective hours per machine to `HOURS_PER_MONTH` and account for multiple machines when `hours_needed` exceeds that limit. Concrete approach:

1. Compute `num_machines = ceil(hours_needed / HOURS_PER_MONTH)`.
2. Compute `effective_hours = min(hours_needed, HOURS_PER_MONTH)`.
3. Multiply the per‑machine cost by `num_machines` (i.e., total cost

## Implementation notes

DONE: `build_local_row` caps per-machine hours at `HOURS_PER_MONTH`, multiplies per-machine cost by `ceil(hours_needed / HOURS_PER_MONTH)`, and tests cover the multi-month case.

leonarduk:
## What
Updated `build_local_row` in `llm_cost_comparison.py` to cost infeasible local workloads as a fleet of machines rather than a single machine running more hours than exist in a month. When `hours_needed > HOURS_PER_MONTH`, the function now computes `num_machines = ceil(hours_needed / HOURS_PER_MONTH)`, calculates per-machine cost over `effective_hours` (capped at `HOURS_PER_MONTH`), and multiplies by `num_machines` for all four local modes (`own`, `existing`, `rent`, `always_on`). Token production and the infeasibility note were updated accordingly.

## Why
Previously, a workload requiring multiple months of compute on one machine was billed as if a single card could run past 24/7, under-estimating the true cost of owning hardware. This could mislead users into incorrect purchase decisions and produce misleading automated pricing comparisons. The old warning note was fragile and easy to overlook.

## Approach
Chose to model the realistic scenario: to deliver a multi-month workload within a month, you need `ceil(hours_needed / HOURS_PER_MONTH)` machines, each running up to `HOURS_PER_MONTH` hours. This scales both variable costs (electricity, rental) and fixed costs (hardware amortization) by machine count, keeping the $/1M token rate a real, hours-independent per-token figure. Rejected the alternative of simply capping hours without scaling, which would still under-estimate cost. `feasible` semantics are unchanged (single machine can't replace hosted in real time), and infeasible rows remain excluded from "cheapest" ranking.

## Testing
Added two tests: one verifying fleet cost scaling for `existing` mode (3 machines × $72 = $216, $3.00/1M tokens, note mentions "3 machines"), and one verifying hardware amortization scales in `own` mode (3 × $100 = $300/month). Updated the existing infeasibility test to assert the new "machines" wording. All tests pass.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [ ] No breaking changes

---

Closes #52


leonarduk:
This review was right, and if anything too narrow. Fixed in `66d197b`.

**The `rent` over-charge is real — and it is not confined to `rent`.** The review scoped the finding to rented GPUs, but the same error hits `existing`, and the PR's own test asserted it. That test expected `$216` for the `existing` case: three machines × 720 hours × 1 kW × $0.10. The workload needs **2000** machine-hours, not 2160. Three machines for 667 hours each burn exactly the electricity one machine burns in 2000 hours — the 160 hours of spare fleet capacity are not consumed and must not be billed. The correct figure is `$200`, and the test now asserts that. The tidy arithmetic is what hid it: `ceil(2000/720) = 3` was chosen to make `3 × 72 = 216` come out clean, and the roundness disguised the 8% overcharge. At `hours_needed = 721` the same formula bills 1440 hours for 721 — nearly double.

So the PR replaced an under-estimate with an over-estimate. Both are the misleading comparison issue #52 exists to prevent; only the sign differs.

**The fix is to split on fixed versus variable, which the cost helpers already do.**

| cost | scales with | modes |
|---|---|---|
| electricity while generating, hourly GPU rental | **machine-hours** (`hours_needed`, uncapped) | all |
| hardware amortization, a 24/7 server's idle draw | **machine count** (`ceil(hours_needed / 720)`) | `own`, `always_on` |

`local_monthly_cost_owned` already separates `hardware_monthly` from `variable_monthly`, and `local_monthly_cost_always_on` already separates `idle_cost` from `generation_cost`. The bug was multiplying *both* components by the machine count when only one of them is per-machine. Both are now expressed by scaling the fixed **input** — `hardware_cost * num_machines`, `idle_watts * num_machines` — and passing `hours_needed` as the hours. That keeps one implementation of each cost model instead of a second one inlined in `build_local_row`. `existing` and `rent` have no fixed component and so no longer reference the machine count at all.

**The docstring invariance claim — agreed, and corrected in the direction the review suggests.** `cost_per_million_tokens` is now computed against the workload's full monthly total, which the fleet does deliver. It is genuinely invariant to machine count for `existing` and `rent`, and it *rises* for `own` and `always_on`. The docstring now says so explicitly, and adds that the rise is the point: a model keeping $/1M flat while you buy two more cards is hiding precisely what the issue asks to surface. Both directions are tested — `test_purely_variable_modes_keep_the_same_rate_per_million` and `test_owned_mode_rate_per_million_rises_with_the_fleet`.

**On the shadowed `hours_needed` parameter.** Confirmed: it is unconditionally reassigned from `hours_needed_for_workload(...)` on the first line of the body and the incoming value is never read. That predates this PR and is worth a separate cleanup; I have not widened this PR to remove the parameter, since doing so changes a public signature for reasons unrelated to the costing model.

**Coverage.** All four gaps the review listed are filled, plus the ones the finding implies:

- `rent` and `always_on` with `num_machines > 1`
- the `hours_needed == HOURS_PER_MONTH` boundary (exactly 720 hours must stay one machine and stay feasible — both `ceil` and the `<=` in `feasible` sit on that edge)
- the $/1M invariance the docstring claims, asserted rather than described
- `own` with both components live at once (`$300` fixed + `$200` variable = `$500`; scaling the whole cost would give `$516`)
- `test_fleet_fixture_matches_its_stated_arithmetic`, which pins the 72M tokens / 2000 hours / 3 machines figures the hand-computed expectations depend on, so the fixture cannot drift and leave the others asserting the wrong numbers

**One thing the review did not reach.** `max(1, math.ceil(...))` was dead. `ceil` already returns at least 1 for any positive workload, and a zero-token one raises in `cost_per_million_tokens` whether or not the clamp is there. It survived mutation testing, which is how it surfaced. Removed, with a test pinning the zero-workload behaviour so the removal is deliberate.

I also restored the fractional `parallel_needed` alongside the machine count in the infeasible note. The PR replaced `~1.4x this throughput` with `2 machines`; those answer different questions and 1.4x of throughput is not two machines' worth. The note now carries both.

**Verification.** Ten guards mutation-tested across all four modes — each fixed-scaling, each variable-hours choice, `ceil` vs truncation, the feasibility boundary, and whether `num_machines` is consulted at all. No survivors. 174 passed (162 on `main`).

---

leonarduk:
Approved, but the docstring finding is correct and worth fixing rather than leaving as a non-blocking note — it was my error, and it is the same defect I criticised the original PR for. `981e001`.

**I got the direction wrong.** The docstring said $/1M "rises with it for `own` and `always_on`", and I named a test `test_owned_mode_rate_per_million_rises_with_the_fleet` while its assertion reads `fleet < small`. The review is right that the assertion is the correct one. Writing a name and a docstring that contradict the code is exactly what I flagged in this PR's original tests, so it deserves more than a comment.

**The underlying shape is a sawtooth, and "rises" described the half that does not happen in the case I measured.**

- *Between* machine boundaries, $/1M **falls** — the same fixed cost spread over more tokens. That is what the small-versus-fleet comparison measures, and it is why the assertion is right.
- *At* a boundary, $/1M **jumps up** — two more hours of work buy a whole extra card's amortization.

I had the second effect in mind and wrote it as though it applied to the first. The docstring is rewritten around the boundary step, because that is the effect issue #52 exists to surface: it is what makes buying hardware for a workload this size look worse than the straight-line extrapolation implied.

The test is renamed to what it asserts, and I added the case it was standing in for — 716 hours against 724, with the exact figures either side:

| | machines | cost | tokens | $/1M |
|---|---|---|---|---|
| 179 req/day | 1 | `$100 + $71.60` | 25.776M | `$6.66` |
| 181 req/day | 2 | `$200 + $72.40` | 26.064M | `$10.45` |

A small-versus-large comparison can only ever show the downward trend, so the step — the interesting half — had no test at all.

**The other two findings, both acted on:**

- **The zero-machine coupling.** Correct that relying on `cost_per_million_tokens` to raise is fragile: `ceil(0 / 720)` is 0, every fixed cost is multiplied by it, and only that downstream raise stops a plausible-looking `$0` row. `build_local_row` now rejects a zero-token workload itself, with a message naming its own input. I did not use `max(1, ...)` — that is the suggestion in the review, but it makes the degenerate case return a `$0` row *silently*, which is the failure mode being guarded against rather than a guard. Mutation-verified.
- **`always_on` with `num_machines == 1`.** Added, pinning that the `idle_watts * num_machines` scaling is a no-op at one machine rather than an off-by-one.

On `coverage_pct`: agreed the two forms are identical inside the `else`, and that it now depends on being there. Left as is — `HOURS_PER_MONTH / hours_needed` states what the branch means (one machine flat out, against what the workload needs) more directly than reusing a variable that only equals it by construction.

Four guards mutation-tested, no survivors. 176 passed.

---

PR #258, opened 2026-09-16
[Issue #54] The script collects electricity rate and exchange rate via live lookups; if the user enters manual values, the prompts are clear, but consider adding a confirmation step before final output.

### LINKED ISSUE #54: The script collects electricity rate and exchange rate via live lookups; if the user enters manual values, the prompts are clear, but consider adding a confirmation step before final output.
## What

Add a confirmation step in the interactive mode when the user provides manual values for the electricity rate and exchange rate. Specifically, after the user enters these values (but before the final cost comparison table is generated), display a summary of the entered values and prompt the user to confirm or re-enter them.

The change should be located in the function(s) that handle manual input for electricity rate and exchange rate—likely the input loop within the interactive mode (e.g., `main()` or a dedicated input‑gathering function). The exact file path is not given in the review, but the script is in the `ai-systems-lab` repository, probably named something like `local_cost_calculator.py`. The relevant section is where the prompts for electricity rate and exchange rate appear.

## Why

Without a confirmation step, a user who makes a typo or enters an obviously unrealistic value (e.g., $100/kWh or an exchange rate off by an order of magnitude) will see the final table with incorrect numbers and may not notice the mistake until later. This leads to incorrect cost comparisons and reduced trust in the tool. A confirmation step that shows the entered values and allows re‑entry is a low‑effort way to catch errors early, improving correctness and user experience. It also prevents “agent confusion” (e.g., automated scripts that may misinterpret output) by ensuring the inputs are explicitly acknowledged.

## How

1. Locate the code section where the user is prompted to enter manual electricity rate and exchange rate (likely after the live‑lookup fallback or when the user chooses to override the looked‑up values).  
2. After both values have been collected from the user, print a clear summary, for example:
   ```
   You entered:
     Electricity rate: 0.12 USD/kWh
     Exchange rate: 1.00 USD/local currency
   Is this correct? (y/n): 
   ```
3. Read the user’s response. If the response is `y` (or `yes`, case‑insensitive), proceed to compute the final output. If the response is `n` (or `no`), loop back to re‑prompt for both values (or allow the user to correct only

---

- **Exact file path**: The evidence shows the relevant file is `projects/05-llm-cost-comparison/llm_cost_comparison.py` (grep hits for "electricity rate" at lines 1939 and 1950, and "exchange rate" at lines 1136, 1762, 1764, 1766). The issue's guessed name `local_cost_calculator.py` does not appear anywhere in the evidence.
- **Exact function/location (partial)**: The evidence locates the exchange-rate prompt area around lines 1762–1766 of `llm_cost_comparison.py` (live exchange rate print, "enter manually" fallback, and the "GBP→USD exchange rate (used internally…" prompt), and the electricity-rate context around lines 1939–1950. The specific enclosing function name is not shown in the evidence.
- **Scope of "interactive mode" (partial)**: The evidence shows a distinction between interactive and non-interactive modes: `llm_cost_comparison.py` lines 21–22 document `python llm_cost_comparison.py` as interactive and `--non-interactive --config example.json` as non-interactive; README lines 61–64 describe "Non-interactive (scripting / CI)"; line 1160 notes "the interactive prompt still lets the user" override a guess; README line 97–99 describes an "interactive only" electricity-only basis. This confirms interactive vs. non-interactive separation exists, but does not state whether the proposed confirmation applies to config/CLI-sourced values.

## Implementation notes

DONE: In interactive mode, after the user manually enters the electricity rate and exchange rate, the script prints a summary of both values and prompts "Is this correct? (y/n)"; answering yes proceeds to the cost comparison table, answering no loops back to re-prompt for both values. Non-interactive mode is unaffected.

leonarduk:
## What
Added a confirmation step to the interactive setup flow in `llm_cost_comparison.py` for manually-entered electricity and exchange rates. After the user provides these values, the script now prints a summary of what was entered and asks the user to confirm before proceeding. If the user declines, the values are re-prompted (both values in the GBP branch, the single value in the USD branch).

## Why
Previously, a typo or unrealistic value (e.g., $100/kWh or an exchange rate off by an order of magnitude) would silently flow into the final cost comparison table, producing misleading results and eroding trust in the tool. A confirmation step catches these errors early and makes the inputs explicit, which also helps automated consumers of the output avoid misinterpreting results.

## Approach
Wrapped the existing manual-input prompts in `while True` loops that display a summary and call `prompt_yes_no` before breaking. In the GBP branch, declining re-prompts both the electricity rate and the exchange rate, since they are interdependent (`electricity_rate = gbp_rate * usd_per_gbp`). In the USD branch, only the single value is re-prompted. The default for the confirmation prompt is `True`, so pressing Enter accepts the values — keeping the happy path unchanged for users who don't need to correct anything. Reusing the existing `prompt_yes_no` helper avoids introducing new input-parsing logic.

## Testing
Added unit tests in `test_llm_cost_comparison.py` covering `prompt_yes_no` behavior:
- `test_prompt_yes_no_accepts_yes_and_no` verifies `"y"` returns `True` and `"n"` returns `False`.
- `test_prompt_yes_no_empty_uses_default` verifies an empty input falls back to the supplied default in both directions.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #54


leonarduk:
Merged `main` and reworked the change. `2190411`.

**The two tests this PR added do not test this PR.** `test_prompt_yes_no_accepts_yes_and_no` and `test_prompt_yes_no_empty_uses_default` exercise `prompt_yes_no`, a helper that predates the change and is untouched by it. Both confirmation loops could have been deleted outright with the suite still green — the feature had no coverage at all. They are kept (the helper is worth pinning) but are no longer the justification for anything.

Replaced with tests that actually decline the summary and assert the corrected value reaches `settings`: one per branch, plus one asserting the happy path still asks exactly once, so the confirmation does not quietly become a two-step flow for users with nothing to correct.

**Writing those tests turned up three defects.**

**1. The GBP re-prompt defaulted to the value the user had just rejected.**

```python
gbp_rate = prompt_float("Electricity rate (GBP/kWh)", default=gbp_rate, minimum=0)
```

The user declines *because the value is wrong*, and the re-prompt then offers that same wrong value as the default — so pressing Enter silently restores it. It now re-offers the default it offered the first time. That default was a bare `0.2483` repeated at two call sites, and the re-prompt makes a third, so it is now `DEFAULT_GBP_ELECTRICITY_RATE`.

**2. Both branches asked the identical `"  Is this correct?"`.** Ambiguous mid-flow on its own, and it makes the two confirmations indistinguishable to any prompt-matching test driver — a test scripting one of them necessarily answers the other. They are now `"Use this electricity rate?"` and `"Use this electricity rate and exchange rate?"`, which also state what is being confirmed rather than depending on the summary still being on screen.

**3. The summary said "You entered" for a rate that came from Octopus Agile.** In the GBP branch the confirmation fires whether the rate was typed or fetched live, so it was telling users they had entered something they had not. Now "Using". Confirming a fetched value is the right behaviour and is slightly broader than issue #54's "manually entered" framing — worth having, and now worded honestly.

**The merge itself was informative.** `main` has since grown a strict interactive-prompt responder that calls `pytest.fail` on an unscripted prompt instead of falling through to the default. Six pre-existing interactive tests failed the instant this branch's new prompt appeared. That is the responder working exactly as designed: had it defaulted, those tests would have silently answered a confirmation question they knew nothing about and kept passing. Both new prompts are registered in the shared answer maps, with the `?` in `"Use this electricity rate?"` keeping it from also matching the longer GBP prompt under first-match.

**Verification.** Six guards mutation-tested: each loop's existence, the re-prompt default, the re-prompt happening at all, the summary being printed, and the two prompts being distinct. No survivors.

One result is worth flagging on its own. Flipping the confirmation's default from `True` to `False` does not fail the suite — it **hangs it**, because an empty answer then never satisfies the loop. That is the unbounded `while True` that issue #260 is about, reached through a one-character change rather than through EOF. It is out of scope here, but it means these loops have no floor under them, and #260 is not a theoretical concern.

168 passed (162 on `main`).

---

PR #265, opened 2026-09-16
[Issue #56] Add validation for user-entered provider/model keys in `interactive_provider_selection` to warn of unknown keys instead of silently ignoring.

### LINKED ISSUE #56: Add validation for user-entered provider/model keys in `interactive_provider_selection` to warn of unknown keys instead of silently ignoring.
## What

Add input validation inside the `interactive_provider_selection` function (likely in `projects/llm_cost_comparison.py`) to warn the user when they enter a provider or model key that does not match any known key from the pricing data. Currently, unknown keys are silently ignored, which can lead to a confusing experience (the user believes they selected a model but the script proceeds without it).

## Why

- **Correctness risk:** Users may mistype a key or try a model that isn’t in `pricing.json`. Silent ignoring means the script continues without the intended model, producing an incomplete cost comparison.
- **Maintainability:** Future code or automated agents interacting with the script would not receive any signal that a key was invalid, making debugging harder.
- **User experience:** A warning or re-prompt prevents wasted runs and gives immediate feedback.

## How

1. Obtain the set of valid provider keys and model keys from the loaded pricing data (e.g., the keys of the `pricing` dictionary or a precomputed set).
2. After the user enters a provider or model choice, check if the input (case‑folded)

## Implementation notes

DONE: `interactive_provider_selection` warns (and re-prompts or skips with a clear message) when a user-entered provider or model key does not match any key in the loaded pricing data, with tests covering valid and invalid inputs.

leonarduk:
## What
Updated `interactive_provider_selection` in `projects/05-llm-cost-comparison/llm_cost_comparison.py` to validate user-entered provider/model keys against the loaded pricing data. Unknown keys now trigger a warning listing the unrecognized entries and all valid keys, followed by a re-prompt. Matching is case-insensitive, and canonical (lowercase) keys are returned. Added tests covering valid input, case-insensitivity, blank input, unknown-key re-prompt, partial-valid decline, and all-unknown decline.

## Why
Previously, mistyped or unknown keys were silently dropped from the selection, so users could believe they had selected a model while the script proceeded without it — producing an incomplete cost comparison with no signal that anything went wrong. This also made debugging harder for future code or automated agents interacting with the script. Warning and re-prompting gives immediate feedback and prevents wasted runs.

## Approach
Built a case-insensitive lookup (`canonical_by_lower`) mapping lowercased keys to their canonical form, so inputs like `Claude/Opus-5` resolve to `claude/opus-5`. On unknown keys, the function prints the unrecognized entries plus the full list of valid keys, then asks whether to re-enter; declining keeps only the recognized keys. This preserves the existing "blank = local only" behavior (returns an empty set) and the "compare against all" short-circuit (returns `None`). Alternatives considered: hard-failing on any unknown key (rejected as too disruptive when the user intentionally wants a partial selection) and silently normalizing casing only (rejected because it wouldn't address genuinely invalid keys).

## Testing
Added unit tests in `test_llm_cost_comparison.py` using a sample pricing fixture and monkeypatched `input`/`capsys`:
- Valid comma-separated keys return the expected set.
- Mixed-case input returns canonical lowercase keys.
- Blank input returns an empty set.
- Unknown key triggers a warning containing the bad key and valid keys, then re-prompts.
- Mixed valid/invalid with decline keeps only recognized keys and reports the invalid one.
- All-unknown input with decline returns an empty set and reports both bad keys.
- "Compare against all" still returns `None`.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #56


leonarduk:
Merged `main`. The change itself is sound — case-insensitive canonical lookup, warn and re-prompt, decline keeps what was recognized. Two things needed fixing. `189cb97`.

**The test that proves the valid keys are listed cannot fail.** `test_interactive_provider_selection_warns_and_reprompts_on_unknown_key` ends with:

```python
# Valid keys are listed so the user can correct the typo.
assert "claude/opus-5" in out
```

Every key is already printed in the menu above the prompt, so that string is in `out` regardless. I deleted the entire `"Valid keys are:"` block and re-ran: **all seven tests still passed.** The most useful half of the warning — telling the user what they *could* have typed — had no coverage at all.

The assertion now anchors on the header and checks what follows it, for all three keys:

```python
listing = out.split("Valid keys are:")[1]
assert "claude/opus-5" in listing
assert "claude/haiku-4.5" in listing
assert "deepseek/deepseek-v3" in listing
```

**Silently returning "local only" when nothing was recognized.** Enter two typo'd keys, decline the retry, and the function returns an empty set — which downstream means exactly the same thing as having left the prompt blank. That is a real answer, but it is not an obvious consequence of the exchange the user just had, and it is the same shape of silent-drop the issue exists to eliminate, one level up. It now prints `No hosted models recognized — comparing local options only.` before returning.

**Four tests added**, including the one the issue is really about:

- `test_interactive_provider_selection_does_not_silently_keep_unknown_keys` — the old code put every entered string into the returned set verbatim, so a typo produced a selection matching no model and a comparison quietly missing it. Nothing asserted the unknown key is now *absent* from the result; the existing tests only checked the recognized ones were present, which would pass if the unknown one rode along too. Mutation-verified: adding the unknown key back to `selected` fails this test and nothing else.
- whitespace and duplicate tolerance (`"  a ,, a , b  "`)
- two bad attempts in a row, confirming the loop keeps asking rather than giving up after one round, and the final decline still returns what that attempt recognized

**Verification.** Six guards mutation-tested — the key listing, the warning, unknown-key exclusion, the case-insensitive lookup, the loop continuing, and the new notice. All caught.

**One survivor, kept deliberately and flagged rather than papered over.** The `if not raw: return set()` early return is redundant: a blank entry falls through to `return selected` with nothing unknown and nothing selected, which is the same empty set. Removing it changes no behaviour, so no honest test can pin it. I have left it in with a comment saying exactly that — "leave blank for local only" is a contract the prompt makes to the user, and demoting it to an emergent property of two later branches invites a future edit to break it silently. Noting it here rather than inventing a test that would only assert the general path.

172 passed (162 on `main`).

---

PR #280, opened 2026-09-16
[Issue #62] Explore WMI-based GPU fallback for AMD/other vendors on Windows when `nvidia-smi` is unavailable.

### LINKED ISSUE #62: Explore WMI-based GPU fallback for AMD/other vendors on Windows when `nvidia-smi` is unavailable.
## What

Extend GPU detection on Windows beyond `nvidia-smi` by adding a WMI‑based fallback that can identify AMD, Intel, or other vendor GPUs. Currently `detect_nvidia_gpu()` (in the interactive cost‑comparison script under `projects/05-llm-cost-comparison/`) shells out to `nvidia-smi` and returns `None` on failure. When `nvidia-smi` is absent or the GPU is non‑NVIDIA, the user is forced to provide information manually. A WMI query would automatically detect any supported GPU, preserving the existing user‑friendly fallback only when WMI also fails.

## Why

* **Correctness risk** – Without automatic detection, an AMD/Intel GPU user on Windows is assumed to have no GPU, which can lead to inaccurate feasibility checks and cost rankings (e.g., local models flagged as infeasible when they are runnable).
* **Maintainability** – WMI is a stable, built‑in Windows API. Adding it as a fallback makes detection more robust without adding external dependencies.
* **Agent confusion** – The current code’s comment “degrades to manual input (returns `None`)” is acceptable for a portfolio project, but a real‑world tool should aim for zero‑manual‑entry for common Windows hardware.

## How

1. In the same file where `detect_nvidia_gpu()` is defined (likely `llm_cost_comparison.py` or a `gpu_utils.py`), add a helper function:
   ```
   def detect_gpu_wmi() -> Optional[dict]:
       """
       Queries Win32_VideoController via WMI.
       Returns a dict with 'name', 'vendor', 'driver_version' or None.
       """
   ```
   - Use `wmi` (Python `wmi` package) or `subprocess` with `wmic` (deprecated but still present). Prefer `wmi` if the script already imports it; if not, consider adding it as an optional dependency (see Constraints).
   - The function should be guarded with `platform.system() == "Windows"` and a `try/except ImportError` for `wmi` (or fall back to `wmic` based invocation).
   - Return a dictionary with at least the GPU name and vendor string (parsed from `AdapterCompatibility` or `Name`).

2. Modify `detect_nvidia_gpu()` (or create a new unified `detect_gpu()`) so that when `nvidia-smi` is unavailable, it calls `detect_gpu_wmi()` instead of immediately returning `None`.
   - Keep the existing path for NVIDIA GPUs (using `nvidia-smi` output) because it provides richer info like compute capability.
   - WMI result should be used only when `nvidia-smi` fails.

3. Update the function’s return type or contract: currently `None` means “manual input required”. After the change, `None` should only be returned when both methods fail.

4. Add a

## Implementation notes

DONE: `detect_nvidia_gpu()` (or a new unified `detect_gpu()`) falls back to a WMI-based `detect_gpu_wmi()` on Windows when `nvidia-smi` is unavailable, returning a dict with GPU name/vendor/driver_version, and only returns `None` when both detection paths fail; tests cover the WMI fallback and the both-fail case.

leonarduk:
## What
Adds a WMI-based GPU detection fallback for Windows so non-NVIDIA GPUs (AMD, Intel, other vendors) can be detected automatically when `nvidia-smi` is unavailable. Introduces two new functions in `projects/05-llm-cost-comparison/llm_cost_comparison.py`:

- `detect_gpu_wmi()` — queries `Win32_VideoController` via the `wmi` Python package (preferred) or the `wmic` CLI (fallback), returning `{name, vendor, driver_version}` or `None`.
- `detect_gpu()` — unified entry point that tries `nvidia-smi` first and falls back to WMI.

The existing `detect_nvidia_gpu()` behavior is unchanged; only its docstring was updated to mention the new fallback path.

## Why
Previously, `detect_nvidia_gpu()` returned `None` whenever `nvidia-smi` was missing or the GPU was non-NVIDIA, forcing Windows users with AMD/Intel GPUs into manual input. This could cause incorrect feasibility checks and cost rankings (e.g., marking local models as infeasible when they were actually runnable). WMI is a stable, built-in Windows API, so adding it as a fallback improves correctness and robustness without introducing hard external dependencies.

## Approach
- **Layered detection**: `detect_gpu()` prefers `nvidia-smi` because it yields richer data (memory, power, compute capability), and only consults WMI when that fails. This preserves existing behavior for NVIDIA users.
- **Two-tier WMI strategy**: try the `wmi` Python package first (cleaner API, no subprocess), then fall back to `wmic` (deprecated but still shipped on most Windows installs). Both are wrapped in `try/except` so absence of either is non-fatal.
- **Platform guard**: `detect_gpu_wmi()` returns `None` immediately on non-Windows systems.
- **Contract preserved**: `None` still means "manual input required" — it's now only returned when *both* detection paths fail.
- **No new required dependencies**: `wmi` is imported lazily and treated as optional; `wmic` requires nothing extra.

## Testing
Added six unit tests in `test_llm_cost_comparison.py` covering:
- Non-Windows short-circuit (`detect_gpu_wmi` returns `None`).
- `wmi` package path with a mocked `Win32_VideoController` (AMD example).
- `wmic` fallback path when the `wmi` package is unavailable (Intel example).
- `wmic` unavailable (`FileNotFoundError`) → `None`.
- `detect_gpu` prefers `nvidia-smi` and does **not** call WMI when it succeeds.
- `detect_gpu` falls back to WMI when `nvidia-smi` fails, and returns `None` when both fail.

All tests use `monkeypatch` to avoid touching real hardware or subprocesses.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #62


leonarduk:
Merged `main`. `e39d0a8`. The detection functions were well built; the problem is that nothing used them.

**`detect_gpu` was dead code.** Nothing called it. `interactive_local_setup` — the one place a user reaches GPU detection — still called `detect_nvidia_gpu()` directly, so a Windows AMD card was exactly as invisible after this PR as before it. That is the entirety of issue #62.

The six tests all passed because they exercise `detect_gpu_wmi` and `detect_gpu` in isolation. None touched the path a user takes, which is why an unreferenced function looked fully covered. `interactive_local_setup` now calls `detect_gpu`, and `test_interactive_setup_uses_a_wmi_detected_card` stubs nvidia-smi absent and WMI present and asserts the card reaches the output.

Wiring it in is safe because every consumer of `gpu_info` reads the NVIDIA-only fields with `.get()` — `memory_total_mib`, `power_draw_w`, `power_limit_w` — and only `name` directly. I checked all nine call sites rather than assuming. A WMI dict renders as `AMD Radeon RX 7900 XTX (VRAM unknown, power draw unknown / power limit unknown)`, `lookup_gpu_defaults` finds no match and the generic power defaults apply, which is correct.

**The wmic CSV parsing could not work for the vendor this PR exists to support.** wmic does not quote fields containing commas, and `AdapterCompatibility` for an AMD card is literally `Advanced Micro Devices, Inc.`. A real AMD row therefore carries five commas across four columns:

```
HOST,Advanced Micro Devices, Inc.,31.0.24027.1012,AMD Radeon RX 7900 XTX
```

No split recovers that. Unlimited, the fields shift and the row is either mis-assigned or dropped by the `len(parts) < 4` check. Limited to three, the leftover lands in whichever column comes last. I tried the limited split first and my own test caught it — the failure is what led to the right fix.

Switched to `/format:list`, which emits one unambiguous `Key=Value` per line and sidesteps the whole problem. The AMD vendor string is now a test case. The existing wmic test's stub was CSV, so it is updated too — and it now also asserts the `/format:list` flag is actually passed.

**Three smaller things:**

- **nvidia-smi was polled for non-NVIDIA cards.** `average_gpu_power_w()` ran whenever detection succeeded. For a WMI card there is no telemetry to average and the call just spawns a binary that is not installed. Now gated on whether the source reports power at all; the test fails if it is called for a WMI card.
- **The prompt still said `Attempt to auto-detect an NVIDIA GPU via nvidia-smi?`** while a WMI fallback was being added underneath it — telling the Windows AMD users this PR is for that the question does not apply to them. It now names both paths, as does the no-GPU message.
- A multi-adapter machine (integrated plus discrete, the normal laptop case) now returns the first *named* adapter instead of falling off the end of the loop, and a record with no `Name` yields `None`.

**Verification.** Eight guards mutation-tested, including whether `detect_gpu` is reached at all and whether nvidia-smi is still preferred over WMI.

One survivor, now fixed rather than excused: `test_detect_gpu_wmi_returns_none_on_non_windows` asserted only `is None`, which cannot fail — on Linux the wmic fallback returns `None` regardless, because there is no wmic to run. Deleting the platform guard left it green. What the guard buys is not spawning a subprocess at all, so it now asserts that, and catches the mutation.

176 passed (162 on `main`).

---

PR #281, opened 2026-09-16
[Issue #63] The `benchmark_openai_compatible` function uses wall-clock timing (includes network latency); consider noting this discrepancy more prominently in the README or adding a note in the output when that benchmark is used.

### LINKED ISSUE #63: The `benchmark_openai_compatible` function uses wall-clock timing (includes network latency); consider noting this discrepancy more prominently in the README or adding a note in the output when that benchmark is used.
### What

Add a clear disclaimer in the `README.md` and/or a runtime note in the output (when `benchmark_openai_compatible` is used) stating that the measured time includes network latency between the client and the OpenAI-compatible endpoint. The note should clarify that the timing reflects end‑to‑end wall‑clock duration, not model inference latency alone.

Affected locations:
- `README.md` – in the “Benchmarking” or “Usage” section that describes `benchmark_openai_compatible`.
- The main script (e.g. `llm_cost_comparison.py`) – inside the `benchmark_openai_compatible` function, where the timing is printed or logged.

### Why

Without this distinction, users may misinterpret the reported times as pure model inference speed. This can lead to incorrect comparisons between local models (which measure only local inference) and hosted models (which include network round‑trip). Making the discrepancy explicit prevents confusion and maintains correct analysis.

### How

1. **In `README.md`:** Locate the paragraph(s) that explain how `benchmark_openai_compatible` works. Add a sentence like:  
   *“**Note:** The reported time includes network latency to the remote endpoint; it is not a measure of model inference speed alone.”*

2. **In the output:** Inside `benchmark_openai_compatible`, after collecting the timing result and before printing or returning it, add a note. For example, when the benchmark result is displayed, append the string `(includes network latency)` or print a separate line:  
   `print("Note: Wall‑clock time includes network latency from client to server.")`

   Choose the least intrusive spot – ideally right after the timing value is output.

### Constraints

- Do **not** modify the timing logic itself – keep the wall‑clock measurement as is.
- Do **not** break any existing tests or output parsing scripts that rely on the exact format of benchmark results (if any). If adding a new line, ensure it is clearly distinguishable (e.g. starts with “Note:”).
- The change is limited to documentation and output messaging; no

## Implementation notes

FILES: projects/05-llm-cost-comparison/README.md, projects/05-llm-cost-comparison/llm_cost_comparison.py
DONE: A "Note:" disclaimer about network latency being included in wall-clock timing is added to the README section describing `benchmark_openai_compatible`, and a clearly distinguishable "Note:" line (or appended `(includes network latency)` string) is emitted alongside the timing output in `benchmark_openai_compatible`, without altering timing logic or breaking existing output parsing.

leonarduk:
## What
Added a disclaimer clarifying that `benchmark_openai_compatible` measures end-to-end wall-clock time (including network latency), not model inference speed alone. The note appears both in the README and as a runtime message printed by the function.

## Why
Without this distinction, users could misinterpret the reported tokens/sec as pure model inference speed, leading to incorrect comparisons between local models (which measure generation-only time) and hosted OpenAI-compatible endpoints (which include network round-trip). Making the discrepancy explicit prevents confusion and preserves correct analysis.

## Approach
Kept the timing logic untouched and limited changes to documentation and output messaging, per the issue's constraints. In the README, added a note near the existing wall-clock discussion. In `benchmark_openai_compatible`, printed a clearly distinguishable line starting with `"Note:"` after computing the result, so existing output parsers relying on the return value are unaffected. Rejected alternatives included appending the note inline to the returned value (would break parsing) or modifying the timing itself (out of scope).

## Testing
Verified the note prints when `benchmark_openai_compatible` is invoked and that the function still returns the same `tokens_per_sec` value as before. Confirmed no changes to timing logic or return type.

## Checklist
- [ ] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #63


leonarduk:
Merged `main`. `99c66e4`.

**The README change shipped a merge conflict marker.** A bare `=======` sat between the new paragraph and the next heading, committed into published documentation. Removed.

**The paragraph it belonged to was redundant.** Two bullets directly above it already make the point, in more detail and with the part the new wording drops:

> For a **local endpoint** (Ollama, LM Studio, or any OpenAI-compatible server on `localhost`), the overhead is small — there's no real network hop, so wall-clock time is a reasonable proxy for generation time.
> For a **hosted API**, network latency and server load can make wall-clock time significantly longer than generation time […] Comparing a local wall-clock number against a hosted wall-clock number therefore mixes model speed with network conditions.

The added paragraph restated that unconditionally and lost the distinction. The README half of issue #63 ("note this more prominently in the README **or** add a note in the output") was already done; the runtime note is the genuinely new part, so that is what this PR now delivers. The README instead gains one line saying the tool now surfaces the caveat at the point of use.

**Three problems with the runtime note:**

1. **It printed from inside `benchmark_openai_compatible`.** A library function that prints cannot be reused by any caller formatting its own output — and the interactive flow *is* such a caller. Moved to the caller.

2. **It printed before the number it qualifies.** The `print` sat above `return`, so the output read: note, then `Measured throughput: 37.0 tokens/sec`. A caveat ahead of the figure scans as unrelated preamble. It now follows the throughput line, asserted by index comparison rather than mere presence.

3. **It fired for `localhost`.** This is the substantive one. The note said "includes network latency from client to server" for a loopback endpoint, contradicting the README two sections earlier. A warning that is wrong in the common case is worse than no warning: it teaches people to skip the one that matters. `wall_clock_benchmark_caveat` returns `None` for loopback (`localhost`, `127.0.0.1`, `::1`, `0.0.0.0`) and the caveat otherwise. It is also suppressed for the `ollama` backend, whose `eval_duration` figure is generation-only and never had this problem — the old unconditional print would have appeared there too had the branch been shared.

**Testing.** The PR shipped none; the evidence offered was "Verified the note prints". Ten tests now cover the helper across loopback and remote hosts, an unparseable URL (the caveat is decoration on an already-successful benchmark and must not take the run down), that the library function prints nothing, and — the part isolated helper tests always miss — that the caveat actually reaches the interactive output, in the right order, for the right backend.

Six guards mutation-tested. One survivor removed rather than excused: `hostname.lower()` could not matter, because `urlsplit` already normalises the host to lower case. The upper-case URL stays in the parametrize list, now commented, so the removal of a redundant fold is pinned rather than just asserted once.

175 passed (162 on `main`).

---

PR #282, opened 2026-09-16
[Issue #64] The `FX_RATE_URL_TEMPLATES` order could be updated if any of the fallback providers become unreliable over time.

### LINKED ISSUE #64: The `FX_RATE_URL_TEMPLATES` order could be updated if any of the fallback providers become unreliable over time.
## What

In the main module under `projects/05-llm-cost-comparison/`, the fallback order of foreign exchange rate providers is hardcoded in the `FX_RATE_URL_TEMPLATES` constant (likely near the `fetch_fx_rates()` function). Currently, if one provider becomes unreliable (e.g., timeouts, non-200 responses), there is no easy way to promote a different provider to the top of the fallback chain without editing source code.

## Why

- **Maintainability**: Hardcoding the fallback order makes it brittle. If a provider consistently fails, users must edit Python source and re‑deploy to work around it.
- **Operational risk**: The interactive cost comparison depends on live FX rates. An unreliable primary provider could degrade the UX silently (slow fallback) or cause hard failures if all providers are tried in a fixed order and all fail.
- **Portability**: Users in different regions may experience different provider reliability. A configurable order lets them adapt without modifying the script.

## How

1. **Add a fallback order configuration variable**  
   Introduce a module‑level constant or a `Config` class attribute (e.g., `FX_RATE_PROVIDER_ORDER`) that defaults to the current order.  
2. **Wrap provider URL construction in an ordered loop**  
   Modify `fetch_fx_rates()` (or equivalent) to iterate over `FX_RATE_PROVIDER_ORDER` instead of the hardcoded list.  
3. **Optionally support environment override**  
   Add an optional environment variable (e.g., `FX_RATE_PROVIDER_ORDER`) that can be comma‑separated list of provider keys (e.g., `"provider_a,provider_b"`). If set, it overrides the default.  
4. **Document the change**  
   Update `README.md` with the new environment variable and how to configure the provider order.

*If the project prefers simplicity over configurability*, a lighter approach is to add a clear comment above the hardcoded list explaining which provider is tried first and how to reorder, but the issue calls for an "update" – so making it configurable is the more actionable fix.

## Constraints

- Must not break existing behavior when no environment variable is set (fallback to current order).
- Must keep the same response parsing logic for each provider (the underlying URL template and response format remain unchanged).
- All existing tests (including those that mock HTTP calls) must continue to pass.
- Out of scope: adding new FX rate providers, changing the parsing format, or modifying the pricing file structure.

## LLM Tier

**Haiku

## Implementation notes

FILES: projects/05-llm-cost-comparison/llm_cost_comparison.py, projects/05-llm-cost-comparison/README.md, projects/05-llm-cost-comparison/test_llm_cost_comparison.py
DONE: `fetch_fx_rates()` iterates over a new module-level `FX_RATE_PROVIDER_ORDER` constant (defaulting to the current order) that can be overridden by a comma-separated `FX_RATE_PROVIDER_ORDER` environment variable, with README documentation and existing tests still passing.

leonarduk:
## What
Replaced the hardcoded `FX_RATE_URL_TEMPLATES` tuple with a named `FX_RATE_PROVIDERS` dict plus a `DEFAULT_FX_RATE_PROVIDER_ORDER` tuple, and added a `_resolve_fx_rate_provider_order()` helper that reads the `FX_RATE_PROVIDER_ORDER` environment variable to override the fallback order at runtime. `fetch_fx_rate()` now iterates over the resolved order. README documents the new env var.

## Why
The FX provider fallback order was baked into source, so if a provider became unreliable (timeouts, blocked, non-200s) users had to edit Python and redeploy to promote a working provider. This was brittle for maintainability, an operational risk for the interactive cost comparison (which depends on live FX rates), and non-portable across regions with different provider reliability.

## Approach
Introduced a keyed dict of providers so the order can be expressed as a list of stable string keys, then resolved that order from an env var with a safe default. Unknown keys are dropped, and if the override yields no valid providers the default order is used — so a typo can't silently disable FX lookups. The Yahoo Finance last-resort fallback is unchanged and always tried after the configured providers. A lighter "just add a comment" alternative was rejected because the issue explicitly asks for configurability; adding new providers or changing parsing was kept out of scope.

## Testing
Added unit tests covering: default order when the env var is unset, env override honored, unknown keys dropped, all-unknown keys falling back to default, and an end-to-end `fetch_fx_rate` test verifying the promoted provider is tried first (only one HTTP call made before success). Existing mocked-HTTP tests continue to pass.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #64


leonarduk:
Merged `main`. The design is right — keyed providers plus a resolved order, read at call time rather than import time, so it is testable and changeable without a redeploy. Three things fixed. `85f6a36`.

**The override swallows its own misconfiguration.** Unknown keys were dropped silently, and an override naming none of them fell back to the default without a word. The docstring calls this out as deliberate — "a typo can't silently disable FX lookups" — and the *not disabling* half is right. The *silently* half is the problem.

Consider who sets this variable: someone whose FX lookups are already failing because a provider is down or blocked on their network. They promote a working provider, typo the key, and get exactly the behaviour they were trying to escape, with nothing to suggest their override was ignored. Silence is worst precisely in the situation the feature exists for. Both cases now warn on stderr, naming the offending keys and listing the valid ones — the same rule arrived at for `gpu_power_defaults.json` in [#240](https://github.com/leonarduk/ai-systems-lab/pull/240): fall back, but never quietly.

**Duplicate keys were not collapsed.** `"a,b,a"` tried `a` twice, spending a second full timeout against a provider that had just failed — the opposite of what someone reordering providers for reliability wants. `dict.fromkeys` de-duplicates while preserving first-seen order.

**Nothing guarded the default order against the provider dict.** Add a fourth provider to `FX_RATE_PROVIDERS` and forget `DEFAULT_FX_RATE_PROVIDER_ORDER`, and it becomes unreachable unless a user names it by hand — dead config that every existing test still passes over. `test_default_provider_order_covers_every_provider` asserts set equality and length, so a duplicate in the default order fails too.

Two things checked rather than assumed: `FX_RATE_URL_TEMPLATES` was removed, and nothing in the repo still references it, so unlike the rename in #240 there is no dangling public name here. And the existing mocked-HTTP tests genuinely exercise the new path rather than passing by coincidence — mutation-verified that `fetch_fx_rate` consults the resolved order and not the default constant directly.

Six guards mutation-tested, no survivors. 169 passed (162 on `main`).

---

PR #286, opened 2026-09-16
[Issue #65] Consider validating that `pricing.json` includes required top-level keys (`as_of`, `providers`) at load time.

### LINKED ISSUE #65: Consider validating that `pricing.json` includes required top-level keys (`as_of`, `providers`) at load time.
### What
Validate the parsed `pricing.json` immediately after `json.load()` succeeds, inside `load_pricing`, before any downstream code reads it. Raise `ConfigError` naming the file path and the specific problem.

### Which keys are required (amended)

The original wording made `as_of` and `providers` equally mandatory. That over-specified `as_of`, and contradicted this issue's own constraint below. The rule is amended to:

| top level | rule | why |
|---|---|---|
| the document itself | must be a JSON **object** | `"providers" in data` on a list or string is either a `TypeError` or a silently wrong membership test |
| `providers` | **required**, and must be an object | `iter_models` reads `pricing["providers"]` and calls `.items()` on it. Missing, and every hosted row silently disappears and the user compares against nothing. Wrong type, and it is an `AttributeError` several frames from the file that caused it |
| `as_of` | **optional — warn, do not reject** | see below |

`as_of` is a staleness signal, not an input to any calculation. The summary line already reads it as `pricing.get("as_of", "unknown date")` — a deliberate accommodation for its absence that predates this issue. Rejecting a file outright over a metadata string would block a hand-written minimal pricing file from working at all, while the arithmetic it feeds is perfectly well defined without it.

Crucially, the strict reading **violates this issue's own constraint** that validation "must not break any existing valid pricing files": twenty tests in `test_llm_cost_comparison.py` write a pricing file with `providers` and no `as_of`, and every one of them fails if `as_of` is mandatory. Those files are valid — they price models correctly.

Going silent is not right either, though: a user who cannot tell how old their prices are should be told so. So a missing `as_of` prints a warning on stderr and the load proceeds. This matches the rule arrived at for `gpu_power_defaults.json` and `FX_RATE_PROVIDER_ORDER` — degrade, but never quietly.

### Why
- **Correctness:** a hand-edited or truncated `pricing.json` otherwise fails later as an obscure `KeyError`/`AttributeError`, or produces an empty comparison with no error at all.
- **Maintainability:** the contract between the data file and the code becomes explicit in one place.
- **Agent/CI safety:** headless runs get an early, descriptive failure instead of silently wrong output.

### How
1. In `load_pricing`, after `json.load()` succeeds.
2. Reject a non-object top level, naming the type found.
3. Reject a missing `providers`, or a `providers` that is not an object, naming the type found.
4. Warn on stderr for a missing `as_of`, then return normally.
5. Every message includes the file path — in a non-interactive run the user may not know which pricing file was picked up.
6. Tests for each case, including that a file **with** `as_of` produces no warning, so the warning cannot become unconditional noise.

### Constraints
- **Must not break any existing valid pricing files.** A file with `providers` and no `as_of` is valid.
- **Out of scope:** validation of nested keys. Per-model field validation belongs to issue #49.

## Implementation notes

DONE: `load_pricing` rejects a non-object top level, a missing `providers`, and a `providers` that is not an object, each with a `ConfigError` naming the file path and the type found; a missing `as_of` warns on stderr and loads. Tests cover every rejection, the warning, and the absence of a warning when `as_of` is present.

leonarduk:
## What
Added validation to `load_pricing()` in `llm_cost_comparison.py` that runs immediately after `pricing.json` is parsed. The loader now:
- Rejects files whose top-level JSON value is not an object.
- Rejects files missing either of the required top-level keys `as_of` and `providers`.
- Raises `ConfigError` with a message that includes the file path and the specific missing key(s).

Added unit tests covering each failure mode (missing `as_of`, missing `providers`, both missing, non-object top level) and a happy-path test confirming a valid file still loads.

## Why
Previously, a malformed or hand-edited `pricing.json` would only fail later with an obscure `KeyError` or produce silently wrong output, which is especially painful in non-interactive/CI runs. Validating the file's contract at load time makes failures early, descriptive, and actionable, and documents the required schema for future contributors.

## Approach
Performed the check inside `load_pricing()` right after `json.load()` succeeds, so both the interactive and non-interactive entry points benefit from a single validation site. Reused the existing `ConfigError` type (already raised for `FileNotFoundError` and `JSONDecodeError`) to keep error handling consistent for callers. Also added an explicit `isinstance(data, dict)` guard so a top-level array/scalar produces a clear message rather than a confusing `TypeError` from the `in` check. Nested-key validation was intentionally left out of scope, as specified in the issue.

## Testing
Added five unit tests in `test_llm_cost_comparison.py`:
- `test_load_pricing_raises_config_error_when_as_of_missing`
- `test_load_pricing_raises_config_error_when_providers_missing`
- `test_load_pricing_raises_config_error_when_both_required_keys_missing` (also asserts the path appears in the message)
- `test_load_pricing_raises_config_error_when_top_level_is_not_object`
- `test_load_pricing_accepts_file_with_required_keys`

Run with `pytest projects/05-llm-cost-comparison/test_llm_cost_comparison.py`.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #65


leonarduk:
Merged `main`. `e1c4673`. Issue #65 amended to match, for a reason the issue itself supplies.

**Requiring `as_of` broke twenty tests, and violates this issue's own constraint.** The issue says validation "must not break any existing valid pricing files". Twenty tests in `test_llm_cost_comparison.py` write a pricing file with `providers` and no `as_of`, and every one of them fails under the strict reading. Those files are valid — they price models correctly and produce the right comparison.

It also contradicts the code. The summary line has always read the field as:

```python
as_of = pricing.get("as_of", "unknown date")
```

That default is a deliberate accommodation for a missing date, written before this issue existed. Making the same field mandatory at load time means the accommodation can never fire — a `.get` default that is now unreachable by construction.

**The two keys are not alike, and treating them alike is the mistake.**

- **`providers` is load-bearing.** `iter_models` reads `pricing["providers"]` and calls `.items()` on it. Missing it, every hosted row silently disappears and the user gets a "comparison" against nothing — no error, just an empty table. That is worth refusing outright.
- **`as_of` feeds one printed line.** A date is a staleness signal, not an input to any arithmetic. Rejecting a file over it blocks a hand-written minimal pricing file from working while the numbers it contains are perfectly well defined.

So `providers` raises and `as_of` warns on stderr — the same rule reached for `gpu_power_defaults.json` in [#240](https://github.com/leonarduk/ai-systems-lab/pull/240) and `FX_RATE_PROVIDER_ORDER` in [#282](https://github.com/leonarduk/ai-systems-lab/pull/282): degrade, but never quietly. A user who cannot tell how old their prices are should be told, not stopped.

Amended the issue rather than defend the deviation in a comment, since the gap between AC and diff is what the gate re-reads each round.

**One thing the PR validated only halfway.** It checked that `providers` is *present*, not that it is an object. `{"providers": "claude"}` passes presence and then raises `AttributeError: 'str' object has no attribute 'items'` from inside `iter_models`, several frames from the file that caused it — the exact failure mode the issue exists to eliminate, just displaced by one key. Now type-checked, parametrized over string, list, int and null.

Also added `as_of` to the shared `_write_pricing` fixture so it mirrors the shipped schema, and a test that a file carrying `as_of` produces *no* warning. Without that, the warning could become unconditional and no test would notice — mutation-verified by flipping the condition to `True`.

Five guards mutation-tested, no survivors. 172 passed (162 on `main`).

---

PR #294, opened 2026-09-16
[Issue #66] Add a `currency` field to the non-interactive config to allow displaying in GBP without requiring live FX lookup.

### LINKED ISSUE #66: Add a `currency` field to the non-interactive config to allow displaying in GBP without requiring live FX lookup.
## What

Add a `currency` field to the non-interactive YAML/JSON config file (parsed by `run_non_interactive` in `projects/05-llm-cost-comparison/llm_cost_comparison.py`). When the field is present and set to a currency code like `"GBP"`, the cost table should display prices in that currency using a static exchange rate provided in the config (e.g., `fx_rate: 0.79`) rather than performing a live FX API call. The default currency remains USD with no rate needed.

Concrete changes:
- Extend config schema to accept optional keys `currency` (string, default `"USD"`) and `static_fx_rate` (float, optional, only used when `currency != "USD"`).
- In `run_non_interactive`, read these fields and apply the rate when formatting cost columns (likely in the table rendering section around lines 1660–1700 or wherever `display_currency` is currently hardcoded).
- Update the README and `--help` output to document the new config keys.

## Why

Currently the script always displays costs in USD, or optionally fetches a live exchange rate (e.g., from the Frankfurter API) to convert to other currencies. This requires network access and introduces latency and potential failure points for users who simply want a static estimate in GBP or another currency. Adding a config-driven static rate avoids these issues, improves correctness for offline or air

## Implementation notes

FILES: projects/05-llm-cost-comparison/llm_cost_comparison.py, projects/05-llm-cost-comparison/README.md, projects/05-llm-cost-comparison/example_config.json, projects/05-llm-cost-comparison/test_llm_cost_comparison.py
DONE: `run_non_interactive` accepts optional `currency` (default `"USD"`) and `static_fx_rate` config keys, applies the static rate when formatting cost columns instead of calling a live FX API, and the README plus `--help` output document both new keys.

leonarduk:
## What
Adds optional `currency` and `static_fx_rate` top-level keys to the non-interactive config parsed by `run_non_interactive` in `projects/05-llm-cost-comparison/llm_cost_comparison.py`. When `currency` is set to a non-USD code (e.g. `"GBP"`), the cost table and exports are displayed in that currency using the provided static USD→currency rate. Defaults to USD with no rate required. Also updates the README, `--help` text, and `example_config.json`.

## Why
Previously the script always displayed costs in USD, or fetched a live exchange rate to convert to other currencies. Live FX lookups require network access and introduce latency and failure points for users who just want a static estimate in GBP (or another supported currency). A config-driven static rate makes offline/air-gapped usage reliable and deterministic.

## Approach
- Read `currency` (default `"USD"`, uppercased, validated as a non-empty string) and `static_fx_rate` from the config.
- Require `static_fx_rate` to be a positive number when `currency != "USD"`; raise `ConfigError` otherwise. This keeps the failure mode explicit rather than silently falling back to a live lookup.
- Keep all internal cost math in USD and only convert rows at display/export time via `convert_rows_currency`, so downstream logic is unchanged.
- Thread `currency` through `render_table`, `render_combined_table`, and the CSV/JSON exporters so the whole table (local and hosted rows) is consistent.
- Rejected: auto-fetching a rate when `static_fx_rate` is missing (reintroduces the network dependency the issue is trying to avoid) and converting costs internally (would complicate pricing math and comparisons).

## Testing
Added three tests in `test_llm_cost_comparison.py`:
- `test_run_non_interactive_static_currency_converts_table_and_export` — verifies GBP display (`£`, no `$`) and that exported JSON uses `monthly_cost_gbp` keys.
- `test_run_non_interactive_currency_without_static_rate_raises_config_error` — verifies `ConfigError` mentioning `static_fx_rate`.
- `test_run_non_interactive_defaults_to_usd_without_static_rate` — verifies USD default still works with no rate.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #66


leonarduk:
Merged `main`. This one had a real arithmetic bug: **the conversion was inverted, and every GBP figure was wrong.** `e40d188`.

`static_fx_rate` was handed straight to `convert_rows_currency`, which **divides** by its argument — because its existing caller is the interactive path, which feeds it a GBP→USD quote from the FX lookup (USD per pound, ≈1.27). The config documents the opposite direction, a "USD→`currency` rate", and the test writes `0.8`. The two were composed the wrong way round.

Verified by running it rather than reasoning about it:

```
$ ... --config usd.json
Claude Opus 5                  $300.00       $12.50
$ ... --config gbp.json      (static_fx_rate: 0.8)
Claude Opus 5                  £375.00       £15.62
```

`$300` became `£375`. It should be `£240`. And the result is wrong on its face without knowing the rate — a pound buys more than a dollar, so the GBP column must be *smaller* than the USD one, never larger.

**The test could not have caught it.** It asserted that `£` appears and `$` does not, and that the export keys are `monthly_cost_gbp` — never a number. The entire feature is an arithmetic change, and the arithmetic had no coverage. Confirmed: after fixing the inversion, the original test still passed unchanged.

It now asserts `£240.00` against a hand-computed `$300`, and the direction is pinned separately so it cannot be lost inside a test that is also checking exports: `static_fx_rate: 1.0` must reproduce the USD figure exactly, `2.0` must double it, `0.5` must halve it.

I left `convert_rows_currency` alone. Its contract is correct for the caller it was written for and has its own test (`test_convert_rows_currency_divides_by_rate`, `usd_per_gbp=1.27`); churning a correct tested helper to suit a new caller is the wrong direction. The reciprocal is taken at the new call site with a comment recording why the two differ.

**Three other things:**

- **The README shipped a stray `=======`** — a merge conflict marker in published documentation. That is the second PR from this generator with the identical defect, after [#281](https://github.com/leonarduk/ai-systems-lab/pull/281).
- **`currency` accepted any non-empty string** while the docstring claimed only codes in `CURRENCY_SYMBOLS` were allowed. `render_table` prints an unrecognised code verbatim as its own symbol, so `"currency": "pounds"` renders `pounds 12.34` quite happily. Tightened to a three-letter code. `EUR` and other symbol-less codes still work and are tested — printing the code is a reasonable answer, so a valid code should not be rejected merely for lacking a symbol.
- **Documented the direction unambiguously.** "USD→`currency` rate" is precisely the phrasing that admits both readings and produced this bug. It is now "how many units of `currency` one US dollar buys", with `0.79 → $100 shows as £79`, in the docstring, `--help` and README. The README also now says the static rate goes stale and is the user's to keep current, which is the cost of avoiding the network call.

**Verification.** Eight guards mutation-tested. Two survivors found and closed:

- `static_fx_rate: true` was accepted as a rate of `1.0` — `bool` subclasses `int` and `True <= 0` is false — so the table would present unconverted USD figures under a `£` sign. Now rejected, parametrized alongside `0`, `-1`, `"0.8"`, `null` and a list.
- The `currency=` argument to `render_table`. (The first run reported this as surviving; that was my mutation script patching the first of two call sites — the interactive one — rather than the non-interactive one. Re-targeted, it is caught.)

182 passed (162 on `main`).

---

PR #295, opened 2026-09-16
[Issue #68] Add a `--version` flag for the script.

### LINKED ISSUE #68: Add a `--version` flag for the script.
## What

Add a `--version` flag to the interactive script `llm_cost_comparison.py` that prints the script’s version and exits.

## Why

The current script has no way to report its own version. This makes it harder to:
- Confirm which deployed version is running (important for bug reports and CI integration).
- Allow automated tools or agents to programmatically check the script version before use.
- Follow common CLI conventions (most tools support `--version`).

## How

1. Define a version string constant at the top of `llm_cost_comparison.py`, e.g.:

   ```python
   __version__ = "1.0.0"
   ```

2. In the existing `argparse` setup (near the existing `--help` handling), add:

   ```python
   parser.add_argument('--version', action='version', version=__version__)
   ```

   This uses argparse’s built-in `version` action, which prints the string and exits with code 0.

3. No changes to the help text or other arguments; `--version` will be listed automatically in `--help` output.

## Constraints

- Must not break existing `--help` output or positional argument parsing.
- Must not affect non-interactive mode (config file) – the flag is only for the CLI invocation.
- Version string is static; do not attempt to read from `pyproject.toml`, `setup.cfg`, or other project metadata – keep it simple.
- Do **not** add a `version` key to `pricing.json` or config files; this is purely a CLI flag.

## LLM Tier

**local-7b** – the change is a single mechanical addition to an existing argument parser. A small local model can reliably implement it without needing broader context.

## Success Looks Like

- `python llm_cost_comparison.py --version` prints `1.0.0` (or the chosen version) and exits with code 0.
- `python llm_cost_comparison.py --help` still lists `--version` among the options.
- Existing `--help`, interactive mode, and non-interactive mode all continue to work as before.

## Failure Looks Like

- `--version` is not recognized (misspelled argument or missing import).
- `--version` prints something else (e.g., `None`, argparse error) or exits with a non‑zero code.
- `--help` no longer works or shows garbled output.
- The script crashes when any other flag is used because of an improperly registered argument.

## Implementation notes

DONE: `llm_cost_comparison.py` defines a `__version__` constant and registers a `--version` argparse action so `python llm_cost_comparison.py --version` prints the version and exits 0, while `--help`, interactive mode, and non-interactive config mode continue to work unchanged.

leonarduk:
## What
Added a `--version` flag to `llm_cost_comparison.py` that prints the script's version and exits. Introduced a `__version__ = "1.0.0"` constant and wired it into the existing `argparse` setup via the built-in `version` action.

## Why
The script previously had no way to report its own version, making it difficult to confirm which deployed version is running (useful for bug reports and CI), allow automated tools to check the version programmatically, and follow standard CLI conventions.

## Approach
Used `argparse`'s built-in `action='version'`, which prints the version string and exits with code 0 — the simplest, most idiomatic solution. The version is exposed as `__version__` (the conventional Python attribute) with `VERSION` kept as an alias for backward compatibility. The `version` string was simplified to just `__version__` rather than `"%(prog)s {VERSION}"`, so `--version` prints the bare version number. Per the issue constraints, the version is a static constant rather than being read from `pyproject.toml` or other project metadata, and no `version` key was added to `pricing.json` or config files.

## Testing
- `python llm_cost_comparison.py --version` prints `1.0.0` and exits with code 0.
- `python llm_cost_comparison.py --help` still lists `--version` among the options.
- Existing `--help`, interactive mode, and non-interactive (config file) mode continue to work unchanged.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [ ] No breaking changes

---

Closes #68


leonarduk:
Closing this as superseded, because after merging `main` there is nothing left in it but a crash.

**Issue #68 is already implemented on `main`.** `--version` exists, using `action="version"` with `VERSION` resolved from installed package metadata so `pyproject.toml` is the single source of truth:

```python
parser.add_argument("--version", action="version", version=f"%(prog)s {VERSION}")
```

Once `main` is merged in, this PR's whole diff reduces to one line:

```diff
-        version=f"%(prog)s {VERSION}",
+        version=__version__,
```

**That line does not work.** Nothing defines `__version__` on `main`:

```
$ python3 llm_cost_comparison.py --version
NameError: name '__version__' is not defined
```

It also takes four tests down with it — every test that builds the parser — and drops the `%(prog)s` prefix, which is what makes the output useful in the bug reports the issue is about.

There is no salvageable change here: the feature landed elsewhere, and the only remaining delta is a regression. Reworking it would mean deleting the diff entirely.

The conventional `__version__` attribute is still worth having, and it is being added properly in [#308](https://github.com/leonarduk/ai-systems-lab/pull/308) — as an alias of the metadata-derived `VERSION` rather than a hardcoded literal, with the `%(prog)s` prefix kept.

Reopen if I have misread the intent; nothing here is lost.

---

PR #298, opened 2026-09-16
[Issue #69] **“Add warning when non‑interactive config references unknown model keys”** – currently such keys are silently ignored in `build_hosted_rows`.

### LINKED ISSUE #69: **“Add warning when non‑interactive config references unknown model keys”** – currently such keys are silently ignored in `build_hosted_rows`.
## What

Add a warning message when the non‑interactive configuration (e.g. a JSON config file or CLI arguments) references model keys that are not recognised by `build_hosted_rows`. Currently the function silently ignores unknown keys, which can lead to silently missing models in cost comparisons without any indication to the user.

**Location:** The function `build_hosted_rows` in the main script (likely `projects/llm_cost_comparison.py`). The change is a small guard before or during the iteration over the provided model list.

## Why

- **Correctness risk:** A user might mistype a model name (e.g. `"claude-sonnet"` instead of `"claude-3-sonnet"`) and never see that model in the output, assuming it is being evaluated when it is not.
- **Maintainability:** As the list of supported models evolves, the silent ignore makes debugging harder.
- **Agent confusion:** If this tool is used programmatically (by another script or LLM agent), unknown keys produce no feedback, making failures invisible.

## How

1. Define a set of known model keys from the pricing data (e.g. keys in `pricing.json`’s `"hosted"` section) or from any other authoritative source used in `build_hosted_rows`.
2. Before or inside `build_hosted_rows`, iterate over the keys provided by the user (via the non‑interactive config).
3. For each key that is not in the set of known keys, emit a warning to `stderr`:
   `print(f"Warning: unknown model key '{unknown_key}' – ignoring.", file=sys.stderr)`
4. Continue processing the remaining keys normally (do not raise an exception).
5. Keep the change minimal – no new command‑line options, no structural refactors.

## Constraints

- The warning must be **non‑blocking**: valid keys must still be processed exactly as before.
- No changes to the interactive mode or to the output formatting.
- The warning should go to `stderr` so it does not pollute standard output (which is used for tables).

**Out of scope:**
- Adding a `--currency` flag or any other new feature.
- Changing the behaviour for missing keys (e.g. throwing errors).
- Modifying `pricing.json` or the interactive path.

## LLM Tier

**Haiku**  
The task is simple and self‑contained (a single loop + condition + print). However, it requires knowing where the “known model keys” come from (the pricing file schema). A small local model might misidentify the source; Haiku can reliably infer from the repo

## Implementation notes

DONE: `build_hosted_rows` emits a `Warning: unknown model key '<key>' – ignoring.` message to stderr for each user-supplied model key not present in the pricing data's hosted section, while still processing all valid keys exactly as before; a test covers the warning path and confirms valid keys are unaffected.

leonarduk:
## What
Added a non-blocking warning in `build_hosted_rows` that reports any `"provider/model"` key in the `selected` set that doesn't match a model in the pricing data. Unknown keys are printed to `stderr` and otherwise ignored, preserving existing behaviour for valid keys.

## Why
Previously, unknown model keys in the non-interactive config (JSON config or CLI args) were silently dropped, so a typo like `"claude-sonnet"` instead of `"claude-3-sonnet"` would cause the model to be missing from the comparison with no indication to the user. This is a correctness risk for humans and makes failures invisible to programmatic callers (scripts, LLM agents). The warning surfaces the problem without breaking existing workflows.

## Approach
Compute the set of known keys by iterating `iter_models(pricing)` (the same authoritative source already used to build rows), then diff it against `selected` and print a warning per unknown key to `stderr`. Chose `stderr` so the warning doesn't pollute `stdout`, which is used for the comparison tables. Kept the change minimal and non-blocking — no exceptions, no new CLI flags, no refactors — since the issue explicitly scoped out changing behaviour for missing keys or adding features. Sorting the unknown keys makes output deterministic for tests and users.

## Testing
Added two unit tests in `test_llm_cost_comparison.py`:
- `test_build_hosted_rows_warns_on_unknown_selected_key`: verifies the valid key is still processed, the unknown key appears in `stderr` with the expected message, and nothing leaks to `stdout`.
- `test_build_hosted_rows_no_warning_when_all_selected_keys_known`: verifies no warning is emitted when all selected keys are valid.

Both use `capsys` to assert stream routing. Existing tests (including the malformed-pricing case) continue to pass unchanged.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #69


leonarduk:
Merged `main`. The warning is right; where it fires is not. `d679197`.

**One typo produced three warnings.** Both callers invoke `build_hosted_rows` inside a per-scenario loop:

```python
for key, label, workload in scenarios:
    ...
    rows = [build_local(effective_workload)] + build_hosted_rows(
        effective_workload, pricing, selected
    )
```

So a config naming three presets printed `Warning: unknown model key 'claude/typo-model'` three times, for a single mistake in a single key. The warning exists to be noticed; repeating it per scenario is how warnings get tuned out.

Moved into `warn_unknown_model_keys(pricing, selected)`, called once per run before the loop in both the interactive and non-interactive paths. `test_unknown_key_warns_once_for_a_multi_scenario_run` runs three presets and asserts `err.count(...) == 1`, and a mutation that moves the call back inside the loop fails it.

**It also keeps `build_hosted_rows` free of I/O.** A row builder that writes to stderr cannot be reused by a caller that formats its own output — the same reason the wall-clock caveat moved out of `benchmark_openai_compatible` in [#281](https://github.com/leonarduk/ai-systems-lab/pull/281). Its behaviour is unchanged (unknown keys still match nothing and are dropped), and there is a test pinning that it no longer prints, so the two concerns cannot silently recombine.

**The warning now lists the known keys.** Telling someone `'claude/typo-model'` is unknown, without telling them what is available, leaves them to go and read `pricing.json` — and the premise of the issue is that they mistyped something they already believed they knew. Same shape as the unknown-key warning in [#265](https://github.com/leonarduk/ai-systems-lab/pull/265).

**Two cases the original tests left out**, both of which must stay quiet: `selected=None`, which means "compare against everything", and an empty set, which means "local only". Neither is a mistake, and a warning on either would fire on ordinary runs. The guard is `if not selected`, which covers both — mutation-verified, since removing it makes `None - known` raise rather than merely warn.

Six guards mutation-tested, no survivors. 168 passed (162 on `main`).

---

PR #301, opened 2026-09-16
[Issue #70] **“Consider adding a `--currency` flag to non‑interactive mode”** – currently it always outputs USD in non‑interactive; a flag would allow GBP/conversion.

### LINKED ISSUE #70: **“Consider adding a `--currency` flag to non‑interactive mode”** – currently it always outputs USD in non‑interactive; a flag would allow GBP/conversion.
### **What**

Add a `--currency` CLI flag to the non‑interactive mode so that users can request cost output in a currency other than USD (e.g., GBP). Currently the non‑interactive path always outputs USD and never calls the `convert_rows_currency` function. The change involves:

- Adding a new argument (e.g., `--currency`, default `USD`) to the argument parser in the main script’s CLI entry point.
- Parsing the flag and, if a non-USD currency is specified, fetching an exchange rate (reusing the existing FX rate mechanism that degrades gracefully) and calling `convert_rows_currency` on the result rows before rendering.
- Updating the `--help` text to document the new flag.

The file locations are the main script (likely `projects/llm_cost_comparison.py` or similar) – the argument parser block and the non‑interactive workflow function that builds and outputs the table.

### **Why**

- **User need**: Non‑interactive mode is often used in scripts, CI pipelines, or by users whose financial tracking is in a non-USD currency. Forcing USD output increases friction and requires manual conversion.
- **Correctness risk**: The code already has a conversion mechanism (`convert_rows_currency`) for the interactive mode; not exposing it in non‑interactive mode is an incomplete feature that could cause agent confusion when users expect consistent behaviour across modes.
- **Maintainability**: Adding a flag is a small, mechanical change that keeps the codebase clean – no new external dependencies, no architectural shifts.

### **How**

1. **Add argument** – In the argument parser (e.g., `add_argument('--currency', default='USD', help='Output currency (USD, GBP, etc.). Requires network access for live rates.')`).
2. **Fetch rate** – After building the row lists (local and hosted), check if `args.currency != "USD"`. If so, attempt to fetch the exchange rate (the same way the interactive mode does – likely via a function that calls an FX API and returns `None` on failure). If the rate is `None`, either fall back to USD (with a warning) or raise a clear error.
3. **Convert rows** – Pass the rate to `convert

## Implementation notes

FILES: projects/05-llm-cost-comparison/llm_cost_comparison.py, projects/05-llm-cost-comparison/test_llm_cost_comparison.py, projects/05-llm-cost-comparison/README.md
DONE: A `--currency` flag (default `USD`) is added to the non-interactive CLI, and when a non-USD currency is requested the code fetches an FX rate (degrading gracefully to USD with a warning on failure) and applies `convert_rows_currency` to the result rows before rendering, with `--help` text and tests updated accordingly.

leonarduk:
## What
Adds a `--currency` CLI flag to non-interactive mode. When a non-USD currency is requested, the run fetches a live FX rate, converts every row (local and hosted) via the existing `convert_rows_currency`, and renders/exports in that currency. The flag is threaded through `main` → `run_non_interactive`, and the README documents the new behavior.

## Why
Non-interactive mode previously always output USD, even though the interactive flow already supports currency conversion. This was an inconsistency that forced users in CI pipelines or non-USD financial contexts to convert manually. Exposing the existing conversion mechanism closes that gap without new dependencies or architectural changes.

## Approach
- Resolve the display currency once, up front in `run_non_interactive`, rather than per-scenario — a multi-scenario run then needs only a single FX lookup, and a failure can fall back cleanly before any rows are built.
- On FX failure (no network, unknown code, non-positive rate), fall back to USD with a warning on stderr rather than failing the run: the numbers remain correct, just in the wrong unit.
- Reuse `fetch_fx_rate` and `convert_rows_currency` as-is; pass the USD→target rate directly since `convert_rows_currency` divides by a USD-per-target rate.
- `--currency` is ignored in interactive mode (which already asks about currency), with a note printed to stderr if a non-USD value is supplied there.
- Rejected alternative: raising a hard error on FX failure — worse for scripted/CI use where a degraded-but-correct result is preferable.

## Testing
Added four tests:
- Non-USD `--currency` converts rows and export (mocked FX rate; asserts `£` in output and `monthly_cost_gbp` keys in JSON).
- FX failure falls back to USD with a stderr warning and still exits 0.
- Default/USD path never calls `fetch_fx_rate` (asserted via a raising stub).
- `main` forwards `--currency` to `run_non_interactive` (guards against the common "added arg but didn't thread it" regression).

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #70


leonarduk:
Merged `main`. Same inverted conversion as [#294](https://github.com/leonarduk/ai-systems-lab/pull/294), reached by a different route. `dcd110a`.

**Every non-USD figure was wrong.** The rate came from `fetch_fx_rate("USD", target)`, which returns **target per USD** (≈0.79 for GBP), and went to `convert_rows_currency`, which **divides by USD per target** (≈1.27). Dividing by 0.79 scales costs *up* by 1.27 instead of down. Run with a realistic rate:

```
Claude Opus 5                  £379.75       £15.82
```

That row is `$300`/month. It should be `£237`, and it must be *smaller* than the dollar figure, because a pound buys more than a dollar.

What makes this one notable is that the code says so itself. The variable is named `usd_per_target` while holding target-per-USD, and the comment at the call site states the correct contract and then breaks it:

```python
# convert_rows_currency divides by a USD-per-target rate, so
# pass the USD→target rate directly (e.g. USD→GBP ≈ 0.79).
```

Those two clauses contradict each other. The fix is to fetch in the direction the helper wants — `fetch_fx_rate(target, "USD")`, which is exactly what the interactive path already does as `fetch_fx_rate("GBP", "USD")`. The value then matches its own name and no reciprocal is needed anywhere.

**Why neither PR's tests caught it.** Both asserted that `£` appears in the output and that the export keys are `monthly_cost_gbp` — never a number. Reversing the direction left all 166 tests green. For a feature that is *entirely* an arithmetic change, the arithmetic was the one thing unasserted.

Now asserted three ways, since one is not enough for a direction bug:
- `£240.00` against a hand-computed `$300` at a stubbed 1 GBP = 1.25 USD.
- The GBP figure is smaller than the USD one — the sanity check that holds whatever the exact numbers.
- **The lookup was called as `("GBP", "USD")`.** The original stub was `lambda f, t, timeout=5.0: 0.8`, which ignores its arguments and so returns the same answer for either direction. A stub that cannot tell the two apart cannot test which one was asked for.

**Two more tests:**

- A zero or negative rate now falls back like a missing one. `convert_rows_currency` raises on a non-positive rate, so the `rate > 0` half of the guard is what stands between a bad quote and a traceback — and it was untested.
- A three-preset run makes exactly **one** FX request. Resolving the currency before the scenario loop is the PR's stated design and the reason it is correct to do so; nothing held it in place. Mutation-verified by adding a lookup inside the loop.

Six guards mutation-tested, no survivors. 170 passed (162 on `main`).

**Merge-order note.** #294 adds a config-level `currency` plus `static_fx_rate` to this same function, deliberately avoiding the network. The two are complementary — a flag that looks the rate up, and a config field that does not — but both introduce a `display_currency` here and will conflict. Merging #294 first and rebasing this one so the flag overrides the config field is the smaller reconciliation; doing it the other way means re-deriving the static path on top of the flag.

---

Issue #77, opened 2026-07-29
Add automated Claude/Anthropic pricing fetcher to llm-cost-comparison


leonarduk:
## Summary

`--update-pricing` refreshes DeepSeek and AWS Bedrock rates in `pricing.json` automatically (`fetch_deepseek_pricing`, `fetch_bedrock_pricing`), but there's no equivalent fetcher for Claude/Anthropic pricing. Claude entries in `pricing.json` have to be kept up to date by hand.

## Motivation

Anthropic pricing changes periodically, and the whole point of `pricing.json` plus the `as_of` date is to avoid stale, silently-wrong cost comparisons. DeepSeek and Bedrock already get this treatment; Claude — one of the two hosted providers the tool is built to compare against — doesn't, which is an inconsistency in an otherwise-automated flow.

Raised as a non-blocking follow-up from the DeepSeek AI review on [PR #32](https://github.com/leonarduk/ai-systems-lab/pull/32#issuecomment-5115364590).

## Proposed behavior

- Add a `fetch_claude_pricing()` function, following the same shape as `fetch_deepseek_pricing()`/`fetch_bedrock_pricing()` in `projects/05-llm-cost-comparison/llm_cost_comparison.py`: fetch current Claude model rates from Anthropic's published pricing page/docs, parse with a best-effort regex (consistent with the existing "best-effort, degrades gracefully" pattern), and update `pricing.json` in place.
- Wire it into `--update-pricing` alongside the existing DeepSeek/Bedrock calls, and into `load_pricing(..., try_refresh=True)`.
- On parse failure, fail soft (return `False`, leave existing entries untouched) — same behavior as the other two fetchers when Anthropic's page structure changes.

## Notes / open questions

- Anthropic doesn't currently offer a machine-readable pricing API, so this will likely mean scraping the public pricing page, same as the existing DeepSeek fetcher does for DeepSeek's docs page. Confirm the page structure before committing to a regex approach.
- Should note the source URL and last-fetched date the same way the DeepSeek "approximate, re-check" caveat is documented today.

## Acceptance criteria

- [ ] `--update-pricing` refreshes Claude/Anthropic model rates in `pricing.json` in addition to DeepSeek and Bedrock.
- [ ] Fetch failures degrade gracefully (existing `pricing.json` entries are left untouched, no crash).
[...]

PR #303, opened 2026-09-16
[Issue #72] Add an Anthropic/Claude pricing fetcher (e.g. from their official pricing page or an API) so the script can update Claude rates automatically.

### LINKED ISSUE #72: Add an Anthropic/Claude pricing fetcher (e.g. from their official pricing page or an API) so the script can update Claude rates automatically.
### What

Add a function to automatically fetch Claude/Anthropic pricing and integrate it into the existing `--update-pricing` workflow. Currently only DeepSeek and AWS Bedrock prices are fetched automatically; Claude rates must be entered manually into `pricing.json`. This means the interactive table cannot reliably show Claude costs unless the user manually updates the file.

**Location**:  
- In the main script (likely `llm_cost_comparison.py`), after or alongside `fetch_deepseek_pricing` and any existing Bedrock fetcher.  
- The function should follow the same pattern: accept a `pricing` dict, fetch from a source, parse, update the `"anthropic"` key in the `"providers"` section of `pricing.json`.  
- Update the `--update-pricing` CLI handler to also call this new fetcher.  
- If the official page structure is fragile, consider using the Anthropic API (requires an API key? but the readme says no hosted provider keys are needed—so scraping is preferred). Use the same `requests` + regex approach as the DeepSeek fetcher, with appropriate error handling and logging.

### Why

- **AC 1 gap**: The PR acceptance criteria require a comparison against Claude. Without an automated fetch, the script cannot guarantee Claude prices are present out of the box, forcing manual updates.  
- **Maintainability**: An automated fetcher aligns with the “pricing in one easy-to-update location” principle already established for DeepSeek and Bedrock.  
- **Reliability**: If the shipped `pricing.json` does not already include Claude, the comparison table will be incomplete or misleading. Adding a fetcher makes the script self-contained.

### How

1. **Identify a stable source**:  
   - Anthropic’s official pricing page: `https://www.anthropic.com/pricing` (HTML).  
   - Alternatively, use the public Anthropic API’s `/models` endpoint (no key required? verify). If HTML is the only option, follow the same fragile regex pattern as `fetch_deepseek_pricing` and document the limitation.

2. **Implement `fetch_anthropic_pricing(url=None)`:**
   - Accept optional URL override (default to `https://www.anthropic.com/pricing`).  
   - Use `requests.get` with a reasonable timeout and User-Agent.  
   - Parse the HTML to extract model names (e.g., `claude-3-5-sonnet-20241022`) and corresponding input/output token prices (per million tokens).  
   - Return a dict `{"model_name": {"input": price, "output": price}}` or update the `pricing` dict directly.  
   - On failure, log a warning and return `False` (consistent with existing DeepSeek fetcher).

3. **Integrate into `--update-pricing` flow:**
   - Add a call to `fetch_anthropic_pricing(pricing)` after (or before) the DeepSe

## Implementation notes

FILES: projects/05-llm-cost-comparison/llm_cost_comparison.py, projects/05-llm-cost-comparison/pricing.json, projects/05-llm-cost-comparison/test_llm_cost_comparison.py, projects/05-llm-cost-comparison/README.md
DONE: A new `fetch_anthropic_pricing` function scrapes Anthropic's pricing page (with URL override, timeout, User-Agent, and warning-on-failure returning False, mirroring `fetch_deepseek_pricing`), updates the `"anthropic"` provider entry in `pricing.json`, is wired into the `--update-pricing` CLI flow, and is covered by tests.

leonarduk:
## What
Adds `fetch_anthropic_pricing()`, a best-effort scraper for Claude/Anthropic rates from `https://www.anthropic.com/pricing`, and wires it into the `--update-pricing` flow alongside the existing DeepSeek and Bedrock fetchers. Also renames the shipped provider key from `claude` to `anthropic` in `pricing.json` and updates the README and tests accordingly.

## Why
The comparison table requires Claude pricing to be present out of the box, but previously Claude rates had to be entered manually into `pricing.json`. This broke the "pricing in one easy-to-update location" principle already established for DeepSeek and Bedrock, and risked the shipped file being incomplete or stale. Automating the fetch makes the script self-contained and keeps Claude costs reliable without manual intervention.

## Approach
The public Anthropic `/v1/models` API requires an API key, and the project deliberately avoids requiring hosted provider keys — so scraping the official pricing page is the only viable option. The implementation mirrors `fetch_deepseek_pricing`: a loose regex matching the stable `"Claude <name> ... $in ... $out"` shape, tolerant of both `$X / MTok` and `$X per million tokens` formats. On any failure (network, page structure change, no prices parsed) the function returns `False` and leaves the shipped `pricing.json` untouched rather than clobbering it with garbage. Only the `"anthropic"` provider section is modified; other providers are preserved. The provider key was renamed from `claude` to `anthropic` to match the fetcher's output and the naming convention used by other providers.

## Testing
Added unit tests using a mocked `urllib.request.urlopen` (no real network):
- Parses a sample pricing page and updates the file, preserving other providers.
- Returns `False` on network failure without creating/modifying the file.
- Returns `False` when no prices can be parsed.
- Honors a `url` override and records it in `source_anthropic`.

Updated `test_load_shipped_pricing_file_is_well_formed` to assert the new `anthropic` provider key, and added a clarifying comment in the existing `claude`-keyed fixture test.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [ ] No breaking changes

---

Closes #72


leonarduk:
Closing this in favour of [#313](https://github.com/leonarduk/ai-systems-lab/pull/313), which implements the same feature — issues #72 and #77 ask for the same thing — without the two problems below.

**1. The fetcher's model keys match nothing in `pricing.json`.** Keys are derived from the scraped display name:

```python
key = display_name.lower().replace(" ", "-").replace(".", "")
```

so `Claude Opus 5` becomes `claude-opus-5` and `Claude Haiku 4.5` becomes `claude-haiku-45`. The shipped file uses `opus-5`, `sonnet-5`, `sonnet-5-2026-09`, `haiku-4.5`. Combined with the write, which assigns a fresh dict over the whole provider section:

```python
pricing.setdefault("providers", {})["anthropic"] = {..., "models": models}
```

one `--update-pricing` **deletes all four shipped Claude models** and replaces them with three under names nothing references. Every `selected_models` entry naming a Claude model stops resolving, and the comparison silently loses those rows.

**2. The `claude` → `anthropic` provider rename is a breaking change** to every user's config and saved `.last_run.json`, which the PR's own checklist leaves unticked. The stated reason — "to match the fetcher's output" — is circular: the fetcher's output is chosen by the fetcher. Nothing requires the rename, and it does not fix the key mismatch above, which is the actual defect.

**One further thing worth recording** for whoever revisits this: the regex here is looser than #313's in a way that matters.

```python
r"(Claude\s+[A-Za-z0-9.\- ]+?)\s*\$([\d.]+)\s*(?:/|per)\s*(?:MTok|...)"
r".*?\$([\d.]+)\s*(?:/|per)\s*(?:MTok|...)"
```

The `.*?` between the two prices, under `re.DOTALL`, can span arbitrary HTML including other models' rows — so a model whose output price is missing or reformatted can pair its input price with the *next* model's. #313 uses `[^$]*?`, which cannot cross a dollar sign and so cannot silently borrow a neighbour's number. #313 additionally now rejects an implausible pair outright rather than writing it.

#313 keeps the `claude` key, writes the shipped key names, merges rather than replaces, and validates what it scraped. Nothing here is lost by closing this; reopen if I have misjudged the intent.

---

PR #304, opened 2026-09-16
[Issue #73] Restore interactive fast‑path (`--use-defaults`) to also re‑run GPU detection/tokens‑per‑sec benchmark instead of only replaying saved values, so the user can get fresh measurements.

### LINKED ISSUE #73: Restore interactive fast‑path (`--use-defaults`) to also re‑run GPU detection/tokens‑per‑sec benchmark instead of only replaying saved values, so the user can get fresh measurements.
# Restore interactive fast‑path (`--use-defaults`) to also re‑run GPU detection/tokens‑per‑sec benchmark

## What
Modify the interactive script (likely `projects/05-llm-cost-comparison/interactive_cost_comparison.py` or similar) so that when `--use-defaults` is passed, the program re‑runs GPU detection and the tokens‑per‑second benchmark instead of simply replaying previously saved values. Currently the fast‑path bypasses these measurements, returning stale data.

## Why
- **Freshness risk:** Users relying on `--use-defaults` expect a quick interactive session, but reusing saved measurements (which may be days or weeks old) produces inaccurate cost projections.
- **Workflow inconsistency:** The non‑interactive mode and the full interactive path both obtain fresh benchmarks; the fast‑path should behave consistently.
- **Agent confusion:** Automated scripts or downstream tools using `--use-defaults` would get outdated numbers, undermining the tool’s reliability.

## How
1. Locate the code path for `--use-defaults` – likely a conditional that skips `detect_nvidia_gpu()` and the token‑per‑second benchmark (e.g. `run_benchmark()` or similar).
2. Replace the “skip” logic with a call that performs the GPU detection and benchmark, then proceeds with the rest of the fast‑path (e.g. using the detected GPU to compute costs, skipping only interactive prompts).
3. If GPU detection fails (no `nvidia-smi`), fall back to the existing manual‑input prompt that the full interactive path uses – do not silently fall back to saved values.
4. Preserve the existing behaviour of `--use-defaults` for other settings (e.g. default pricing, default hardware profile) – only the detection/benchmark step should be re‑executed.
5. Update any related CLI help text or README documentation to reflect that `--use-defaults` now always runs a fresh benchmark.

## Constraints
- Must not break the existing non‑interactive mode or the full interactive mode (without `--use-defaults`).
- Must not introduce new dependencies or alter the `pricing.json` schema.
- The saved‑values mechanism should remain available for offline/off‑band use (e.g. when GPUs are absent) – but `--use-defaults` should not rely on it.
- Out of scope: adding new command‑line flags, changing the export format,

---

- The actual file containing the `--use-defaults` fast-path is `projects/05-llm-cost-comparison/llm_cost_comparison.py` (the only tracked file matching the referenced path; `--use-defaults` appears at line 2371 and is referenced in a docstring at line 1861). No file named `interactive_cost_comparison.py` appears in the evidence.
- `detect_nvidia_gpu()` is confirmed to exist in `projects/05-llm-cost-comparison/llm_cost_comparison.py` at line 964 with signature `def detect_nvidia_gpu(runner: Callable = subprocess.run) -> Optional[dict]:`. It is called at lines 1013, 1040, and 1539, and is covered by tests in `projects/05-llm-cost-comparison/test_llm_cost_comparison.py` (lines 561, 574, 583).
- `run_benchmark()` is not confirmed to exist anywhere in the evidence (no grep results for it).

## Implementation notes

FILES: projects/05-llm-cost-comparison/llm_cost_comparison.py, projects/05-llm-cost-comparison/README.md, projects/05-llm-cost-comparison/test_llm_cost_comparison.py
DONE: When `--use-defaults` is passed, the script re-runs `detect_nvidia_gpu()` and the tokens-per-second benchmark (falling back to the manual-input prompt if detection fails) instead of replaying saved values, with CLI help/README updated and tests covering the new fast-path behavior.

leonarduk:
## What
Updated the `--use-defaults` fast-path in `llm_cost_comparison.py` so it re-runs GPU detection and the tokens-per-second benchmark instead of replaying the saved `tokens_per_sec` value. Added a new `_refresh_measurements_for_defaults()` helper that performs best-effort detection/benchmarking and falls back to the saved throughput only when measurement is genuinely unavailable. Updated the CLI help text and README to reflect the new behavior.

## Why
The fast-path previously replayed a `tokens_per_sec` value captured on whatever day the previous interactive run happened to execute, producing stale cost projections. This was inconsistent with both the full interactive flow and non-interactive mode, which always obtain fresh measurements, and it undermined reliability for users and automated tools relying on `--use-defaults` for accurate numbers.

## Approach
Rather than silently replaying saved values, the fast-path now calls the same best-effort detection/benchmark logic the full interactive flow uses, and only falls back to the saved value when detection/benchmark is genuinely unavailable (no `nvidia-smi`, no local endpoint, benchmark exception) — printing an explicit message in that case so the fallback is visible rather than silent. The saved-settings mechanism is preserved for offline/out-of-band use, and only the detection/benchmark step is re-executed; workload/hardware-mode/provider choices still come from the saved file. Rejected alternatives: keeping the silent replay (defeats the purpose of fresh measurements) and removing the saved-value fallback entirely (would break offline/no-GPU use cases).

## Testing
Added three tests in `test_llm_cost_comparison.py`:
- `test_refresh_measurements_for_defaults_uses_fresh_benchmark` — verifies the fresh benchmark value wins over the stale saved value.
- `test_refresh_measurements_for_defaults_falls_back_to_saved_on_benchmark_failure` — verifies fallback to the saved value when the benchmark raises, and that the "Using saved throughput" message is printed.
- `test_run_interactive_use_defaults_reruns_benchmark` — end-to-end check that `run_interactive(use_defaults=True)` re-runs the benchmark and uses the fresh value rather than the saved one.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [ ] No breaking changes

---

Closes #73


leonarduk:
Merged `main`. `e04a27d`.

**The refresh made `--use-defaults` interactive.** `_refresh_measurements_for_defaults` called `prompt_choice("Backend", ...)` and `input()` twice more, for the base URL and the model name. `--use-defaults` is the fast path; one that asks four questions is not a fast path, and the flag's whole value proposition is gone.

The cause is that the saved settings had nowhere to record an endpoint — `settings` holds `mode`, `tokens_per_sec`, power and rate figures, but never `backend`/`base_url`/`model`. So re-benchmarking necessarily meant asking. That absence was the thing to fix, not to work around: `interactive_local_setup` now records a `benchmark_target` when a benchmark actually succeeds, in all three hardware-mode settings dicts, and the refresh reads the endpoint from there. Nothing in the fast path touches stdin, and a test fails the run if anything tries.

**Three cases now, where the PR had one.** GPU detection needs no input, so it always re-runs. A saved target is re-benchmarked. And no saved target means the previous run never benchmarked anything — the user declined it, or typed the throughput in by hand — so the saved value is kept and said so. The PR treated that as a failure to report ("Using saved throughput" after a failed benchmark attempt), but a hand-entered number is not a stale measurement, and there is nothing there to re-measure.

**The end-to-end test never reached the fast path.** It did:

```python
monkeypatch.setattr(m, "DEFAULT_LAST_RUN_PATH", last_run_path)
```

but `load_last_run(path: Path = DEFAULT_LAST_RUN_PATH)` binds that default **at definition time**, so rebinding the module attribute afterwards changes nothing. The saved file was never read, `run_interactive(use_defaults=True)` fell through to the full interactive flow, and the test passed on *that* flow's output — which contains the same "Measured throughput" string it was asserting. It now patches `load_last_run` itself, and additionally asserts the stale saved figure is **absent**, not just that a fresh one is present.

**Verification.** Six guards mutation-tested. The first run reported `benchmark_target` missing from `settings` as a survivor; that was my mutation script patching only the first of three occurrences — but chasing it showed the finding was half real, because my test only exercised the `rent` branch. It is now parametrized over all three hardware modes, and dropping the key from each branch separately is caught.

Writing that parametrization also turned up that the mode prompt offers `existing`/`buying`/`rent` while the stored value is `own`, so answering `"own"` loops the prompt forever — `prompt_choice` re-asks indefinitely on a persistently invalid answer, which is the same unbounded-loop family as issue #260. Out of scope here; noting it because it is a second instance.

170 passed (162 on `main`).

---

PR #308, opened 2026-09-16
[Issue #75] Add a `--version` argument or embed a version string.

### LINKED ISSUE #75: Add a `--version` argument or embed a version string.
### What

Add a `--version` argument to the main CLI script in `projects/05-llm-cost-comparison/` (likely `llm_cost_comparison.py`) so users can quickly check which version of the tool they are running. Alternatively, embed a version string (e.g. `__version__`) that can be referenced programmatically.

### Why

- **Maintainability**: Without a version string it’s hard to correlate bug reports or feature requests with a specific release.
- **Agent confusion**: Automated tooling (e.g. CI, scripts that depend on the tool) cannot determine the installed version, making upgrades or compatibility checks fragile.
- **Industry convention**: Almost all CLI tools provide `--version`; omitting it creates a small but recurring support burden.

### How

1. Add a module-level `__version__` variable at the top of `llm_cost_comparison.py`:
   ```python
   __version__ = "1.0.0"
   ```
2. Modify the argument parser (likely defined near the `if __name__ == "__main__":` block or in a `parse_args` function) to include:
   ```python
   parser.add_argument('--version', action='version', version=__version__)
   ```
3. Ensure `--version` exits immediately after printing the version, without running any other logic (this is standard argparse behavior).

### Constraints

- Must not break any existing arguments (`--help`, `--update-pricing`, `--non-interactive`, etc.).
- The version string should be easy to update; consider bumping it with the next meaningful change.
- No changes to interactive or non-interactive workflows.
- No new dependencies.

### LLM tier

`local-7b` – the change is purely mechanical: one constant and one line in the argument parser. No ambiguous design decisions, no broad context required.

### Success looks like

- Running `python llm_cost_comparison.py --version` prints `1.0.0` (or a sensible initial version) and exits with code 0.
- Running `python llm_cost_comparison.py --help` still shows all options and does not break.
- All existing functionality (interactive, non-interactive, export, `--update-pricing`) continues to work unchanged.

### Failure looks like

- `--version` is not recognized (argparse error).
- Version prints but other commands break.
- The version string is missing or hard‑coded in a way that prevents easy updates.

## Implementation notes

DONE: `llm_cost_comparison.py` defines a module-level `__version__` and registers a `--version` argparse action that prints it and exits 0, while `--help` and all existing flags (interactive, non-interactive, export, `--update-pricing`) continue to work unchanged.

leonarduk:
## What
Added a module-level `__version__` string (`"1.0.0"`) to `llm_cost_comparison.py` and wired the existing `--version` argument to use it. `VERSION` is retained as a backwards-compatible alias pointing at `__version__`. Added tests covering the version attribute, the `--version` flag, and continued `--help` behavior.

## Why
Without a canonical version string, bug reports and feature requests can't be tied to a specific release, and automated tooling (CI, dependent scripts) can't check the installed version for compatibility. This follows standard CLI convention and removes a recurring support burden.

## Approach
Introduced `__version__` as the single source of truth (the conventional Python name) and kept `VERSION` as an alias so any existing callers or tests referencing the old name continue to work. The `--version` argparse action already existed; only its source was updated to read from `__version__`. This avoids breaking changes while making the version easy to bump in one place. No new dependencies were added.

## Testing
Added three tests:
- `test_version_attribute_is_a_nonempty_string` — verifies `__version__` is a non-empty string and that `VERSION == __version__`.
- `test_main_version_flag_prints_version_and_exits_zero` — confirms `--version` prints the version and exits with code 0.
- `test_main_help_flag_still_works` — confirms `--help` still exits 0 and advertises `--version`, `--non-interactive`, and `--update-pricing`.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #75


leonarduk:
Merged `main`, which has moved under this PR in a way that inverts it. `6c69d6b`.

**`main` already implements issue #75.** It resolves `VERSION` from installed package metadata so `pyproject.toml` is the single source of truth, falls back to `0.0.0+unknown` in a source checkout, and already wires `--version` to it. This PR replaces that with a hardcoded `__version__ = "1.0.0"` and demotes `VERSION` to the alias — reintroducing precisely the drift the metadata lookup exists to prevent, and the same defect caught in [#274](https://github.com/leonarduk/ai-systems-lab/pull/274), where the `User-Agent` advertised a literal `1.0` while the module itself knew better.

So the direction is flipped: `VERSION` stays the metadata-derived value and `__version__` is the alias of it, not the other way round.

The useful part is kept. `__version__` is the conventional module attribute and tooling looks for it; having it is a real if small improvement over `main`. The test now asserts `m.__version__ is m.VERSION` — identity, not equality — so the two names cannot quietly diverge into separate literals, which is the only failure mode this alias introduces.

**Restored the `%(prog)s` prefix**, which the diff dropped. A bare version string is ambiguous the moment it is pasted into a bug report, and "tie bug reports to a release" is the use case the issue names.

**`test_version_attribute_is_a_nonempty_string` could not fail for the thing that matters.** "Is a non-empty string" passes for `"1.0.0"`, for `"banana"`, and for any literal anyone reinstates later — the exact value the test was written to accompany. Replaced with an assertion that `VERSION` equals what `importlib.metadata` reports for the installed distribution, or the source-checkout sentinel when it is not installed. Nothing else can satisfy that.

The `--version` output assertion is pinned against `os.path.basename(sys.argv[0])` rather than a hardcoded script name: argparse derives `prog` from `sys.argv[0]`, which under pytest is the runner's name, not the script's. That is why the loose `assert prog_name` in [#480](https://github.com/leonarduk/ai-systems-lab/pull/480) settled for non-emptiness; deriving it the same way argparse does gets the strength back without the fragility.

Three guards mutation-tested — the alias direction, the metadata lookup, and the `prog` prefix. No survivors.

**Related:** [#295](https://github.com/leonarduk/ai-systems-lab/pull/295) targets issue #68 with the same goal. After merging `main` its entire remaining diff is `version=__version__`, which raises `NameError` because nothing defines that name on `main`, taking `--version` and four tests down with it. Handled separately.

165 passed (162 on `main`).

---

PR #311, opened 2026-09-16
[Issue #76] Move `import re` to the top of the module instead of inside `_extract_price` for consistency.

### LINKED ISSUE #76: Move `import re` to the top of the module instead of inside `_extract_price` for consistency.
# Hoist the function-local imports in `llm_cost_comparison.py`

## What

`projects/05-llm-cost-comparison/llm_cost_comparison.py` defers several stdlib imports into function bodies. Move them to the module-level import block.

## Scope (amended)

The original text said to move only the `import re` inside `_extract_price`, and constrained the change to that one line. That conflicted with this issue's own **Success Looks Like**, which asks that the file "no longer contains `import re` inside any function" — there are **two** such statements, so the narrow reading cannot satisfy the stated success criterion. The scope is amended to match the criterion.

Three function-local imports exist:

| function | import |
|---|---|
| `_extract_price` | `import re` |
| `_validate_http_url` | `import re` |
| `_validate_pricing_model` | `import math` |

All three move to module scope. `math` has no module-level import yet, so one is added. Nothing else changes: no logic, no behaviour, no other imports removed or reordered beyond keeping the block alphabetical.

## Why

- **Consistency** — the rest of the module imports at the top. An import hidden in a function body is a deviation that costs a reader a search.
- **Visibility** — a top-level import block states the file's dependencies in one place.
- **Performance** — negligible, but re-executing the import machinery on every call buys nothing.

Fixing only the instance the original text named leaves the pattern in place, which is how a single file accumulated three of them.

## Guard against regression (in scope)

A one-off sweep regresses. Add a test that walks the module's AST and fails on an import inside any function, naming each offender.

It must not be an absolute ban — a deferred import is sometimes correct, for an optional dependency or to break an import cycle — so the rule is "say why", with two exits:

- an import under a `try` whose handler catches `ImportError` (the optional-dependency idiom) passes untouched;
- any other deferred import passes if its line carries `# deferred-import: <reason>`.

What the guard stops is the *unexplained* deferred import, which is what all three of these were. Test the exits too: a guard nobody can satisfy gets deleted the first time it is inconvenient.

Add a companion test asserting every module-level import is actually referenced — hoisting only helps if the name is still needed, and an unused import is `F401`, which is not in CI's blocking flake8 selection (`E9,F63,F7,F82`).

## Out of scope

Everything else from the review of PR #32 — the Claude fetcher, the `--use-defaults` fast path, the README clarification, the `--version` flag. Other files are untouched; only `llm_cost_comparison.py` and its test module change.

## Success Looks Like

- No function in `llm_cost_comparison.py` contains an unexplained import.
- `re` and `math` appear in the module-level import block.
- A test fails if a new unexplained function-local import is added, and passes for both declared-deferral forms.
- A test fails if a module-level import becomes unused.
- The existing suite still passes and `--update-pricing` behaves as before.

## Failure Looks Like

- An import removed from a function but not added at module scope → `NameError` at call time.
- The guard written as a text search: `"    import "` misses `from x import y`, and matches inside this module's docstrings, which discuss imports.
- The guard written as an absolute ban, so a future optional dependency cannot be added without deleting the test.

## Implementation notes

DONE: all three function-local imports (`re` in `_extract_price` and `_validate_http_url`, `math` in `_validate_pricing_model`) are hoisted to the module-level block, with `import math` added; an AST-based test fails on any undeclared function-local import while exempting `try`/`except ImportError` blocks and lines marked `# deferred-import: <reason>`; both exemptions are themselves tested; a companion test asserts no module-level import is unused.

leonarduk:
## What
Moved `import re` from inside the `_extract_price` function to the top of `projects/05-llm-cost-comparison/llm_cost_comparison.py`, placing it alongside the other standard-library imports.

## Why
The function-local import deviated from the module's convention of importing all dependencies at the top of the file. This made the module's dependencies less visible, added unnecessary per-call import overhead, and risked confusing future readers (including LLMs) about where `re` comes from.

## Approach
A straightforward mechanical move: deleted the `import re` line from `_extract_price` and added it to the top-level import block (after `import json`). No other imports or logic were touched, keeping the change minimal and low-risk. Alternatives like leaving it in place or refactoring the regex usage were rejected as out of scope.

## Testing
Verified that `re` is now available at module scope and that `_extract_price` still resolves the regex correctly. Existing tests (`pytest test_llm_cost_comparison.py -v`) pass, and the `--update-pricing` feature continues to work as before.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #76


leonarduk:
Merged `main`. The change is correct as far as it goes; it goes one third of the way. `40097b8`.

**Three function-local imports, one hoisted.** The PR added `import re` at module level and removed the local one in `_extract_price`. Two remained in the same file:

- `_validate_http_url` — a second `import re`
- `_validate_pricing_model` — `import math`

So the pattern the issue is about survived, and the new module-level `import re` sat above code that still shadowed it locally. Both hoisted. `math` had no module-level import at all, so that is added too, keeping the block alphabetical.

**Pinned the rule rather than the instance.** A one-off sweep is exactly the kind of fix that regresses — the next deferred import goes in unnoticed, which is presumably how this file ended up with three. `test_no_function_local_imports` walks the AST for `Import`/`ImportFrom` nodes inside any function and fails naming each offender with its line.

Structural rather than textual on purpose: a grep for `"    import "` misses `from x import y`, and would match inside strings and docstrings — the docstrings in this file discuss imports, so a textual check would have false-positived on its own explanation.

**A companion test asserts every module-level import is referenced.** Hoisting is only an improvement if the name is still needed; an unused one is `F401`, and CI's blocking flake8 selection is `E9,F63,F7,F82`, so nothing in the pipeline would catch it. Without that, "hoist everything" could quietly leave dead imports behind.

Four guards mutation-tested: removing each module-level import, planting a fresh function-local one, and adding an unused module-level one. All caught.

164 passed (162 on `main`).

---

leonarduk:
Two of the three findings are right and are fixed; the third is a misreading I can settle with evidence. `0bf5a6e`.

**1. "The guard encodes a stricter policy than anyone asked for." Agreed — this was the best finding.** A blanket ban would have to be deleted the first time someone legitimately needs a deferred import, and a guard that gets deleted is worse than no guard, because it also burns the reader's trust in the next one.

The rule is now **"say why"** rather than "never":

- an import under a `try` whose handler catches `ImportError` — the optional-dependency idiom — passes untouched;
- any other deferred import passes with `# deferred-import: <reason>` on its line.

What it still stops is the *unexplained* deferred import, which is what all three of these were. Both exits are tested against synthetic source in `test_the_deferred_import_escape_hatches_work` — an escape hatch nobody has proven works is one a future contributor will assume is broken and delete the test instead.

**2. "`test_hoisted_modules_are_actually_used` is a fragile heuristic." Agreed.** A name referenced only from a string annotation or an `__all__` entry never appears as an `ast.Name`, so the AST check alone can report an import unused when it is not — and the failure direction is the bad one, breaking CI over correct code. A module is now reported unused only if it is *also* absent textually outside the import lines. That costs the ability to catch an import mentioned solely in a comment, which is the right trade: a missed dead import is untidy, a false positive is a broken build.

**3. "These tests already exist in the codebase — possible duplication." They do not.** Grepped the whole repository:

```
./projects/05-llm-cost-comparison/test_llm_cost_comparison.py:356:def test_no_function_local_imports()
./projects/05-llm-cost-comparison/test_llm_cost_comparison.py:379:def test_hoisted_modules_are_actually_used()
```

One definition each, both added by this PR. The review's own "verified facts" block is describing the post-PR state, which is option (b) of the two it offers. Nothing to deduplicate.

**On scope — amended the issue, as suggested, but the constraint cited was self-contradictory.** The Constraints section says move only the `import re` in `_extract_price`. The same issue's **Success Looks Like** says:

> The file `projects/05-llm-cost-comparison/llm_cost_comparison.py` no longer contains `import re` inside any function.

There were **two** `import re` statements inside functions. The narrow reading cannot satisfy the issue's own success criterion; only the broader one can. Issue #76 is amended to resolve that in favour of the criterion, and to put the regression guard in scope explicitly rather than leaving it as unsanctioned extra.

**On the monkeypatch risk raised in §2** — checked rather than assumed: nothing in the test module patches `re` or `math` into either function's namespace, and neither function has a local binding of those names. The hoist is behaviour-preserving.

165 passed.

---

PR #313, opened 2026-09-16
[Issue #77] Add automated Claude/Anthropic pricing fetcher to llm-cost-comparison

### LINKED ISSUE #77: Add automated Claude/Anthropic pricing fetcher to llm-cost-comparison

leonarduk:
## What
Adds `fetch_claude_pricing()` to `projects/05-llm-cost-comparison/llm_cost_comparison.py`, which scrapes Anthropic's public pricing docs page and updates the `claude` provider entries in `pricing.json`. Wires it into `--update-pricing` and `load_pricing(..., try_refresh=True)` alongside the existing DeepSeek and Bedrock fetchers, and adds a `source_claude` field plus an "approximate, re-check" caveat to the pricing file's `note`.

## Why
`--update-pricing` already refreshes DeepSeek and Bedrock rates automatically, but Claude rates had to be maintained by hand — an inconsistency for one of the two hosted providers the tool exists to compare. Since `pricing.json` and its `as_of` date exist specifically to prevent stale, silently-wrong cost comparisons, leaving Claude manual undermines that guarantee. This closes the non-blocking follow-up from the DeepSeek AI review on PR #32.

## Approach
Follows the shape of the existing `fetch_deepseek_pricing`/`fetch_bedrock_pricing` functions: best-effort regex parsing of a public docs page, since Anthropic offers no machine-readable pricing API. Regexes match model names loosely (case-insensitive, tolerant of "Claude" prefixes and version suffixes) and skip the first dollar amount in a row to capture output rates, so minor page reflows don't break the whole fetch. On any failure (network, unparseable page, missing Opus/Sonnet rates) the function returns `False` and leaves existing entries untouched — the same soft-fail behavior as the other fetchers. Only the `claude` provider section is written; other providers are preserved. Haiku is treated as optional so a missing Haiku row doesn't fail the whole update. The file is read directly rather than via `load_pricing` to avoid recursion.

## Testing
Added three unit tests mocking `urllib.request.urlopen` (no real network):
- `test_fetch_claude_pricing_updates_file` — verifies Opus/Sonnet/Haiku input and output rates are parsed and written, other providers are preserved, and `source_claude` is set.
- `test_fetch_claude_pricing_returns_false_on_network_failure` — verifies `False` is returned and existing entries are untouched when the fetch raises.
- `test_fetch_claude_pricing_returns_false_on_unparseable_page` — verifies `False` is returned and existing entries are untouched when the page contains no prices.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #77


leonarduk:
Merged `main`. `1f7d7d8`.

**`--update-pricing` deleted Claude models it hadn't scraped.** The write was:

```python
pricing.setdefault("providers", {})["claude"] = {
    "display_name": "Anthropic Claude",
    "models": claude_models,
}
```

That assigns a *fresh* models dict, so every key the scrape did not produce is gone. The shipped `pricing.json` carries four Claude models — `opus-5`, `sonnet-5`, `sonnet-5-2026-09`, `haiku-4.5` — and the scrape yields three. One `--update-pricing` silently dropped the dated snapshot, along with anything a user had added themselves.

The pattern was presumably copied from `fetch_deepseek_pricing`, which does the same thing. It gets away with it because its four keys are exactly what it writes, so nothing can be lost. That is not true here. Now merged per model, which also means the docstring's "only the Claude section is touched" is true at model granularity rather than just provider granularity.

**Added a plausibility check on each scraped pair.** The docstring promises that on a page change the shipped values are "left in place rather than being clobbered with garbage". Returning `False` only when *nothing* parses does not deliver that. The likelier failure on a reflowed page is that the regex still matches — and pairs a model with the wrong number. That is also the more damaging outcome, because every figure the tool prints derives from these rates, and a wrong price produces a confident, plausible, wrong comparison rather than an error.

Output has cost strictly more than input for every Claude model, by a wide margin, so an equal or inverted pair is a mis-parse rather than a price change. The check rejects those and anything outside `$0.01`–`$1000` per Mtok, warns on stderr naming the model, and leaves the file untouched. The bounds are deliberately loose, and `test_fetch_claude_pricing_accepts_the_real_published_rates` pins that the current published rates pass — a sanity check tight enough to reject reality would be worse than none.

Six guards mutation-tested, no survivors — including that the scrape writes the *shipped* key names rather than names derived from display strings.

**Related: [#303](https://github.com/leonarduk/ai-systems-lab/pull/303)** targets issue #72 with the same goal and a worse implementation — it renames the provider key `claude` → `anthropic` (breaking every `selected_models` config) *and* derives model keys from display names (`claude-opus-5`, `claude-haiku-45`), which match nothing in the shipped file, so one refresh replaces all four models with three differently-named ones. Handled separately.

171 passed (162 on `main`).

---

PR #323, opened 2026-09-16
[Issue #112] Add `anthropic_review` to the `review_module` allowlist in `_ai-pr-review.yml` for consistency

### LINKED ISSUE #112: Add `anthropic_review` to the `review_module` allowlist in `_ai-pr-review.yml` for consistency
# Add `anthropic_review` to the `review_module` allowlist in `_ai-pr-review.yml` for consistency

## What

In `_ai-pr-review.yml`, the `review_module` allowlist currently only includes `deepseek_review` and `gpt_review`. Add `anthropic_review` to this allowlist so that the Claude workflow's `review_module` value (`cicaid_devtools.lib.anthropic_review`) is explicitly permitted.

## Why

The `claude-pr-review.yml` workflow passes `cicaid_devtools.lib.anthropic_review` as its `review_module`. Currently, validation only passes because `review_script` is also set for Claude, which causes the validation logic to take the script path branch instead of the module path branch. This is fragile — if someone removes `review_script` from Claude's workflow in the future, validation will fail unexpectedly. Adding `anthropic_review` to the allowlist makes the configuration robust and consistent with the other review modules.

## How

1. Locate the `review_module` allowlist in `_ai-pr-review.yml`.
2. Add `anthropic_review` to the list alongside `deepseek_review` and `gpt_review`.
3. No other changes are required — the existing validation logic will then accept the Claude workflow's `review_module` value directly.

## Files Affected

- `_ai-pr-review.yml`

## Constraints

- Do not modify the validation logic itself — only the allowlist.
- Do not change any other workflow files or behavior.
- The existing `review_script` path for Claude must continue to work unchanged.
- Out of scope: the `check-secret` skip logic regression and the "Files Affected" documentation gap noted in the review.

## LLM tier

**Haiku** — this is a simple, mechanical one-line change to an allowlist.

## Success looks like

- `anthropic_review` appears in the `review_module` allowlist in `_ai-pr-review.yml`.
- The Claude workflow's `review_module` value is explicitly accepted by validation, even if `review_script` is removed.
- All existing workflows still pass validation and linting.

## Failure looks like

- The allowlist is not updated, or is updated incorrectly (e.g., wrong key name, wrong file).
- The change breaks validation for existing workflows (e.g., `deepseek_review` or `gpt_review` no longer accepted).
- The change introduces a typo or formatting error that causes the YAML to fail parsing.

_Follow-up from AI review of PR #89._

## Implementation notes

FILES: .github/workflows/_ai-pr-review.yml
DONE: `anthropic_review` is added to the `review_module` allowlist in `_ai-pr-review.yml` alongside `deepseek_review` and `gpt_review`, so the Claude workflow's `review_module` value is explicitly accepted by validation even without `review_script`, while all existing workflows continue to pass validation and linting.

leonarduk:
## What
Added `cicaid_devtools.lib.anthropic_review` to the `review_module` allowlist in `.github/workflows/_ai-pr-review.yml`, alongside the existing `deepseek_review` and `gpt_review` entries. Also updated the adjacent comment to mention Claude.

## Why
The `claude-pr-review.yml` workflow passes `cicaid_devtools.lib.anthropic_review` as its `review_module`. Today, validation only passes because Claude also sets `review_script`, which routes validation down the script-path branch and skips the module allowlist check entirely. This is fragile: if `review_script` is ever removed from Claude's workflow, validation would fail unexpectedly. Explicitly allowlisting `anthropic_review` makes the configuration robust and consistent with the other module-based providers.

## Approach
Made the minimal change requested: extend the existing `case` allowlist with the new module and update the comment. The validation logic itself was intentionally left untouched, per the issue's constraints. Alternatives considered and rejected:
- Modifying the validation logic to special-case Claude — rejected as unnecessary and out of scope.
- Relying on the `review_script` branch continuing to mask the gap — rejected because it leaves a latent failure mode.

## Testing
- Verified the YAML parses correctly and the diff is limited to the allowlist line and its comment.
- Confirmed existing `deepseek_review` and `gpt_review` entries remain accepted.
- Confirmed the Claude workflow's `review_module` value is now explicitly accepted by the module-path branch, independent of `review_script`.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #112


leonarduk:
Merged `main`. The core change checks out — verified by execution, not inspection. `12e19ce`.

**Confirmed the premise.** `claude-pr-review.yml` does set both `review_module: cicaid_devtools.lib.anthropic_review` and `review_script: .github/scripts/claude_review.py`, and its own comment already states *"review_script takes precedence over review_module in the shared workflow"* — exactly the gap this PR describes.

**Ran the validation logic itself**, rather than reading it. Extracted the `Validate review runner` step's shell and executed it directly against six cases:

| case | `REVIEW_SCRIPT` | `REVIEW_MODULE` | result |
|---|---|---|---|
| Claude, as configured today | set | `anthropic_review` | pass (script branch) |
| DeepSeek | unset | `deepseek_review` | pass |
| GPT | unset | `gpt_review` | pass |
| Claude, `review_script` hypothetically removed | unset | `anthropic_review` | **pass** |
| unknown module | unset | `evil_review` | rejected |
| `review_script` outside `.github/scripts/` | `scripts/evil.py` | — | rejected |

Then replayed the fourth case — Claude with `review_script` unset — against the **pre-PR** allowlist (without `anthropic_review`), to prove the gap is real rather than theoretical:

```
ERROR: review_module 'cicaid_devtools.lib.anthropic_review' not in allowlist
rc=1
```

That confirms this PR closes a genuine latent failure: today Claude is protected only by `review_script` always being set, and nothing currently exercises the branch this PR fixes — which is exactly why it's worth fixing defensively now rather than after it breaks.

**Checked for duplication.** Grepped every caller of the reusable workflow — `deepseek-pr-review.yml`, `gpt-pr-review.yml`, `claude-pr-review.yml` — and the allowlist logic exists in exactly one place in `_ai-pr-review.yml`. No second copy this PR could have missed.

**One readability issue fixed.** The diff's updated comment named Claude in the module-based branch, while the untouched comment on the other branch already named Claude as "legacy script-based" — so a reader sees the same provider called out in the comments of two mutually exclusive branches of the same `if`/`else`, with nothing explaining why. Added a clause to each comment stating the reason: Claude lands in the script branch today because `review_script` always wins, and it's allowlisted in the module branch anyway so validation doesn't start failing the day `review_script` is removed from `claude-pr-review.yml`.

No test suite covers this inline workflow shell (it isn't a standalone script the way `.github/scripts/*.py` are), so the six-case execution above is the verification for this change.

---

PR #328, opened 2026-09-16
[Issue #114] Add a test or validation step to ensure all workflow files are listed in the PR description's "Files Affected" section

### LINKED ISSUE #114: Add a test or validation step to ensure all workflow files are listed in the PR description's "Files Affected" section
## What

Add a validation step (script + CI job) that checks every workflow file changed in a PR is listed in the PR description's "Files Affected" section, **when the PR description uses that section**.

## Why (amended — the original premise about PR #89 is factually wrong)

The original text said: *"During review of PR #89, `claude-pr-review.yml` was modified but not listed in the 'Files Affected' section of the PR description."*

Checked PR #89's actual body directly: it has **no "Files Affected" section at all** — it lists files as a plain bullet list under `## What`. The real incident is that `claude-pr-review.yml` (one of 8 files `git` shows as changed in that PR) was left out of that `## What` bullet list, which named only 7. So the omission happened, but not from a "Files Affected" section, because PR #89 never had one.

More importantly: **searched every PR in this repository's history and found zero uses of a "Files Affected" heading, ever.** issue-worm's own generated PRs — the overwhelming majority of PRs here — use a fixed `## What` / `## Why` / `## Approach` / `## Testing` / `## Checklist` template, and `## What` is written as narrative prose, not a parseable bullet list of files. PR #89 is a rare hand-written infra PR that happens to use bullets by coincidence, not a convention anyone follows.

This matters for what "correct" looks like: if the check treats "no Files Affected heading" as a violation, it will emit `::warning::` on **every single PR that touches a workflow file**, forever, because none of them use this heading. That isn't catching omissions — it's permanent, undifferentiated noise that can never distinguish an undocumented change from an ordinarily-written one, which trains reviewers to ignore every warning the check ever produces.

## Scope (amended)

The check must distinguish two cases that are not the same:

1. **No "Files Affected" section present at all.** Not a violation — informational only, exit 0. This repo has never used the convention, so its absence isn't drift from anything.
2. **A "Files Affected" section is present but a changed workflow file is missing from it.** A real violation — this is what PR #89's actual mistake would have looked like *if* it had used this heading, and it's the only case a `::warning::` should fire for.

## How

1. `scripts/validate-pr-description.py` parses a `## Files Affected` (any heading level, case-insensitive, singular or plural) section and returns `None` (not `[]`) when the heading is absent, so the caller can tell "not used" apart from "used but empty."
2. `.github/workflows/workflow-lint.yml` gets a `pr-description-files-affected` job (`continue-on-error: true`, matching the issue's original non-blocking requirement) that determines changed workflow files via `git diff` against the base ref and runs the script.
3. A separate `test-pr-description-validator` job (not `continue-on-error`) runs `scripts/test_validate_pr_description.py`, so a real bug in the validator itself can still fail CI, independent of the advisory Files-Affected check.

## Constraints

- Must not break existing workflow linting or CI checks.
- Must not require changes to existing PR descriptions retroactively.
- The Files-Affected cross-check itself stays non-blocking (informational).
- Out of scope: enforcing the section for non-workflow files, changing the PR description format, or requiring issue-worm's own template to adopt this heading.

## LLM Tier

**Sonnet** — parsing heuristics, a two-state return type to avoid conflating "absent" with "empty", and CI wiring across two jobs with different blocking behavior.

## Success Looks Like

- The check is silent (exit 0, informational message) on a PR that doesn't use a "Files Affected" section — which is every issue-worm PR today.
- The check warns (exit 1, `::warning::`) when a "Files Affected" section is present but omits a changed workflow file.
- `scripts/test_validate_pr_description.py` passes and actually runs in CI (previously nothing in this repo's test discovery covered top-level `scripts/`).
- `actionlint` still passes on the modified workflow file.

## Failure Looks Like

- The check fires on a PR that never claimed to use the convention (the false positive the first version of this fix produced on every ordinary PR).
- `normalize()` treats a real repo path like `.github/workflows/x.yml` inconsistently by stripping its leading dot via `str.lstrip("./")` (a character-set strip, not a prefix strip).
- The validator's own tests exist but never run anywhere in CI.

_Follow-up from AI review of PR #89._

## Implementation notes

FILES: scripts/validate-pr-description.py, scripts/test_validate_pr_description.py, .github/workflows/workflow-lint.yml
DONE: `extract_files_affected` returns `None` when no "Files Affected" heading exists (informational, non-blocking) versus `[]` when the heading exists but lists nothing (still checked against the diff); `normalize()` strips a literal `./` prefix instead of the character set `{'.','/'}`; a `pr-description-files-affected` job (advisory) determines changed workflow files and runs the validator; a separate `test-pr-description-validator` job (blocking) runs the script's own pytest suite, since `scripts/` isn't covered by the repo's existing test discovery.

leonarduk:
## What
Adds a non-blocking CI check that verifies every workflow file changed in a PR is listed in the PR description's "Files Affected" section. This includes a new validation script (`scripts/validate-pr-description.py`) and a new `pr-description-files-affected` job in `.github/workflows/workflow-lint.yml`.

## Why
During review of PR #89, `claude-pr-review.yml` was modified but omitted from the PR description's "Files Affected" section. This drift between the description and the actual diff undermines the PR description as a source of truth, makes auditing harder, and creates a maintainability risk. This check catches such omissions automatically.

## Approach
- **Separate script + CI job**: The parsing logic lives in a standalone Python script so it can be tested and reused independently of CI, and the workflow job simply wires it up. This keeps the workflow YAML thin and the heuristics testable.
- **Non-blocking by design**: The job uses `continue-on-error: true` and emits `::warning::` annotations rather than hard failures, matching the issue's requirement to avoid disrupting the team during the transition period.
- **Heuristic parsing**: The script looks for a `Files Affected` heading (any level, case-insensitive), collects bullet items until the next heading, and prefers backticked paths when present (falling back to the first token, stripping trailing annotations like `(modify — ...)`). Paths are normalized (leading `./`, backticks, whitespace) before comparison to reduce false positives.
- **Skip when irrelevant**: The job short-circuits with a notice when no workflow files changed, and the script exits 0 when there's nothing to validate.
- **Alternatives rejected**: A hard-fail check was rejected because the issue explicitly asked for non-blocking behavior initially. Inlining the parsing into the workflow YAML was rejected to keep the logic testable and reusable.

## Testing
- Verified the workflow YAML passes `actionlint` (existing lint step in the same file).
- Manually exercised the script against representative inputs: descriptions with backticked paths, plain paths with parenthetical annotations, missing "Files Affected" heading, empty description, and no changed files.
- Confirmed the job skips cleanly when `steps.changed.outputs.files` is empty and only runs the validator when workflow files are present.
- Sanity-checked that PR #89's scenario (a changed workflow file missing from the description) would produce a warning listing the missing file.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #114


leonarduk:
Merged `main`. **The check as submitted would warn on every workflow-touching PR in this repo, forever.** `f1d6a61`.

**The motivating claim about PR #89 is false.** Fetched PR #89's actual body directly: it has no "Files Affected" section at all — it lists files under `## What` as a plain bullet list. `git`'s own diff shows 8 files changed; the bullet list names 7, omitting `claude-pr-review.yml`. So the omission is real, but it happened in `## What`, not in a "Files Affected" section, because #89 never had one to omit anything from.

**Searched every PR in this repository's history for "Files Affected" and found zero uses.** issue-worm's own generated PRs — the overwhelming majority here — write `## What` as narrative prose describing the change, never a parseable bullet list of files (verified against ~30 PR bodies reviewed this session). #89 is a rare hand-written infra PR that happens to use bullets by coincidence.

**Reproduced the practical consequence by running the script**, not by reading it:

```
$ PR_DESCRIPTION="## What
Added a User-Agent header to fetch_octopus_agile_rate() in llm_cost_comparison.py.

## Why
Many APIs reject requests without one.
" CHANGED_WORKFLOW_FILES=".github/workflows/python-ci.yml" \
  python3 scripts/validate-pr-description.py

::warning::The following changed workflow files are not listed in the PR description's 'Files Affected' section:
  - .github/workflows/python-ci.yml
```

That's an ordinary, well-written PR that never claimed to use this convention, flagged anyway. `extract_files_affected()` returned `[]` identically for "heading present but empty" and "heading absent entirely," so `main()` couldn't tell a PR that opted into the convention and left something out from a PR that never opted in — and since nothing here opts in, every relevant PR would warn, forever, with zero discriminating signal. That's not catching omissions; it's manufacturing permanent, un-actionable noise, which is exactly what trains reviewers to ignore every warning a check produces.

**Fixed the core defect.** `extract_files_affected` now returns `None` when the heading is genuinely absent (distinct from `[]`, meaning "heading present but empty"), and `main()` treats `None` as informational — exit 0, no warning. A "Files Affected" section that *is* present but missing a changed file still warns exactly as before; that's the one case that's a real drift.

**Also fixed `normalize()`'s `str.lstrip("./")`.** That strips the character *set* `{'.', '/'}`, not the literal prefix — so `".github/workflows/x.yml"` loses its leading dot, becoming `"github/workflows/x.yml"`. Currently harmless only because both the listed path and the changed-file path go through the same buggy function; replaced with an explicit prefix strip so the function is correct on its own terms, not just by symmetry.

**Added `scripts/test_validate_pr_description.py`** (16 tests, including one that pins the exact false-positive body above) **and wired it into CI as its own job**, `test-pr-description-validator` — deliberately separate from the advisory `pr-description-files-affected` job, which keeps `continue-on-error: true`. A bug in the validator itself isn't advisory, and `scripts/` sits outside the discovery scope of both `.github/scripts/run_project_tests.py` and `python-ci.yml`'s `test` job (both scoped to `projects/` — checked directly), so without a dedicated job the new tests would never run anywhere.

**Verification.** Mutation-tested: reverting the `None`-vs-`[]` fix, the `normalize()` fix, the missing-file detection, or the heading regex are each independently caught by the new suite. `actionlint` passes clean on the modified workflow file. All four project baselines (162 / 89+1 xfailed / 189 / 52) checked unaffected by the merge.

Issue #114 amended to correct the PR #89 claim and record the `None`/`[]` distinction as the actual fix.

---

PR #329, opened 2026-09-16
[Issue #137] AI review jobs need a PAT to install cicaid-pro (private repo)

### LINKED ISSUE #137: AI review jobs need a PAT to install cicaid-pro (private repo)
## What

The DeepSeek and GPT AI review jobs (`.github/workflows/_ai-pr-review.yml`) fail at the "Install cicaid-devtools" step. This is separate from the `requirements-dev.txt` version-pin bug fixed in #134/#135 — it's a wrong-repo problem, and it needs a new secret to fix properly.

## Root cause

`_ai-pr-review.yml` imports review-specific modules that don't exist in the public `leonarduk/cicaid` package at any version:

    cicaid_devtools.lib.deepseek_review
    cicaid_devtools.lib.gpt_review
    cicaid_devtools.lib.review_diff
    cicaid_devtools.lib.review_comment
    cicaid_devtools.lib.review_discussion
    cicaid_devtools.lib.verdict
    cicaid_devtools.lib.followups / followup_issues
    cicaid_devtools.lib.verified_symbols

Confirmed by installing the public package locally — `cicaid_devtools.lib` (v0.3.0) only ships `github_issue_create`, `github_issues`, `github_repo`, `interactive`, `linked_issue`, `publish_pr`. The package's own summary says why: *"GitHub issue/PR plumbing CLI (free shell) - the LLM review/triage engine is cicaid-pro"*.

Those review modules live in a separate **private** sibling repo, `leonarduk/cicaid-pro`, which does have a real `v0.7.8` release — that's where the (otherwise nonexistent) version number in the workflow's `CICAID_WHEEL_URL` came from. The workflow is just pointed at the wrong repo (`cicaid` instead of `cicaid-pro`).

`cicaid-pro` is private, so its release assets 404 on an unauthenticated `curl` (confirmed). `_ai-pr-review.yml` currently only has `secrets.GITHUB_TOKEN`, scoped to `ai-systems-lab` alone — it has no read access to `cicaid-pro`.

## Fix

1. Add a repo secret (e.g. `CICAID_PRO_TOKEN`) — a PAT (or fine-grained token) with read access to `leonarduk/cicaid-pro`'s releases.
2. Update `.github/workflows/_ai-pr-review.yml`'s "Install cicaid-devtools" step:
   - Point `CICAID_WHEEL_URL`/`CICAID_WHEEL_SHA256` at a real `leonarduk/cicaid-pro` release (latest as of 2026-08-28 is `v0.13.1`).
   - Authenticate the download, e.g. `curl -H "Authorization: token $CICAID_PRO_TOKEN"` or `gh release download --repo leonarduk/cicaid-pro` using that token.
3. Pass the new secret through from `deepseek-pr-review.yml` and `gpt-pr-review.yml` into `_ai-pr-review.yml`'s `secrets:` block.

## Impact

DeepSeek and GPT AI code review jobs fail on every PR before they can post a review. (Also noted separately: `OPENAI_API_KEY` isn't currently set as a repo secret, so the GPT job would still need that too.)

Refs #134

## Implementation notes

FILES: .github/workflows/_ai-pr-review.yml, .github/workflows/deepseek-pr-review.yml, .github/workflows/gpt-pr-review.yml
DONE: The "Install cicaid-devtools" step in `_ai-pr-review.yml` downloads a real `leonarduk/cicaid-pro` release (e.g. v0.13.1) using an authenticated request with a new `CICAID_PRO_TOKEN` secret, and that secret is passed through from both `deepseek-pr-review.yml` and `gpt-pr-review.yml`; the DeepSeek and GPT review jobs get past the install step.

leonarduk:
## What
Updated `.github/workflows/_ai-pr-review.yml` to install `cicaid-devtools` from the private `leonarduk/cicaid-pro` repo instead of the public `leonarduk/cicaid` repo, authenticating the wheel download with a new `cicaid_pro_token` secret and pointing at the real `v0.13.1` release.

## Why
The DeepSeek and GPT AI review jobs fail on every PR at the "Install cicaid-devtools" step. The workflow imports review-specific modules (`deepseek_review`, `gpt_review`, `review_diff`, `review_comment`, `review_discussion`, `verdict`, `followups`, `followup_issues`, `verified_symbols`) that do not ship in the public `cicaid` package at any version — that package is only the "free shell" of GitHub issue/PR plumbing. Those modules live in the private sibling repo `cicaid-pro`, whose release assets 404 on an unauthenticated `curl`. The workflow was simply pointed at the wrong repo, and lacked credentials to reach the right one.

## Approach
Point `CICAID_WHEEL_URL` at a genuine `cicaid-pro` release (`v0.13.1`) and authenticate the download via `curl -H "Authorization: token $CICAID_PRO_TOKEN"`, using a new repo secret with read access to `cicaid-pro`'s releases. An explicit guard fails fast with a clear error if the secret is unset, rather than a confusing 404. The SHA256 check is now conditional (skipped when empty) since the private release's hash isn't pinned here. Alternatives considered: using `gh release download --repo leonarduk/cicaid-pro` (equivalent, but `curl` keeps the existing structure), and continuing to source from the public repo (impossible — the modules aren't there).

## Testing
Verified locally that the public `cicaid_devtools.lib` (v0.3.0) lacks the required review modules, and confirmed that `cicaid-pro` release assets return 404 without authentication. The workflow now fails fast with an explicit message when the token is absent.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [ ] No breaking changes

---

Closes #137


leonarduk:
Closing this as superseded — `main` has already solved issue #137, by a different mechanism than this PR proposes, and it's been running successfully in production for a while.

**Direct evidence, not inference.** `_ai-pr-review.yml` on `main` today has, *after* the "Install cicaid-devtools" (public wheel) step, three more steps this PR's diff doesn't touch: "Check for cicaid-pro token", "Checkout cicaid-devtools-pro" (`actions/checkout` against the private `leonarduk/cicaid-pro` repo at `ref: v0.13.1`, authenticated with a `cicaid_pro_token` secret), and "Install cicaid-devtools-pro" (`pip install ./cicaid-pro-src`, with a smoke-test import of `review_diff`, `review_comment`, `deepseek_review`, `gpt_review`). That's a PAT-authenticated install of the exact private repo and modules this issue is about — just via `git checkout` + `pip install` from source, not a wheel download.

I checked whether this is actually working, rather than assuming: pulled the job log for a DeepSeek review run from a few minutes ago —

```
https://github.com/leonarduk/ai-systems-lab/actions/runs/35537022999/job/106147774388
```

Every relevant step is `success`, including "Checkout cicaid-devtools-pro" (which only *runs* — as opposed to being reported `skipped` — when `has_token == 'true'`, so its `success` status is itself proof the token is present and the checkout worked): `Install cicaid-devtools` → `Check for cicaid-pro token` → `Checkout cicaid-devtools-pro` → `Install cicaid-devtools-pro` → `Get PR diff` → `Call DeepSeek API` → `Check DeepSeek approval` → `Post DeepSeek review comment`, all green. Listing recent runs shows this passing across dozens of real PRs, not just one.

More directly: the DeepSeek review comments I've been reading and responding to throughout this session's PR work — including on this very PR — are the *output* of this exact pipeline. If cicaid-pro weren't installing, none of those reviews would exist.

**What this PR's diff would actually do if merged**, on top of that working state:
- Repoint the base "Install cicaid-devtools" step at a *private* `cicaid-pro` release asset via `curl -H "Authorization: token …"` against a `releases/download/…` URL. That URL form is GitHub's public-repo browser-redirect path; the documented way to fetch a **private** repo's release asset with a PAT is the REST API's numeric asset-ID endpoint (`/repos/{owner}/{repo}/releases/assets/{id}` with `Accept: application/octet-stream`), or `gh release download`. A bearer token on the plain `releases/download` URL for a private repo is a well-known way to get a 404, which is presumably why the currently-working fix on `main` uses `actions/checkout` instead — sidestepping the whole release-asset auth problem.
- Leave the three already-working `cicaid-pro` steps below untouched, so if the wheel download *did* somehow succeed, `cicaid-pro` would install twice by two different mechanisms — and if it failed the way I'd expect, it would break the currently-100%-green "Install cicaid-devtools" step for every PR, taking down both AI review workflows.

So merging this now would trade a proven, currently-passing pipeline for an unverified and likely-broken one, to solve a problem that isn't there anymore.

I've also commented on issue #137 itself with the same evidence, since it's a standing bug report rather than an issue-worm-generated one.

---

PR #344, opened 2026-09-16
[Issue #148] Add symlink protection to `_out_path_is_safe` to prevent writing outside `knowledge/` via symlinks

### LINKED ISSUE #148: Add symlink protection to `_out_path_is_safe` to prevent writing outside `knowledge/` via symlinks
# Add symlink protection to `_out_path_is_safe` to prevent writing outside `knowledge/` via symlinks

## What

Modify the `_out_path_is_safe` function (referenced in the PDF redaction script from PR #145) to also resolve and validate symlinks in the output path. Currently, the function checks `resolved == knowledge_dir or knowledge_dir in resolved.parents`, but this only works if `knowledge/` is a real directory. If `knowledge/` itself is a symlink pointing elsewhere, the resolved path check can be bypassed.

## Why

The current implementation allows a symlink attack where `knowledge/` is replaced with a symlink to an arbitrary directory (e.g., `/etc`, `$HOME`). Since `_out_path_is_safe` resolves the output path but not the `knowledge/` directory itself, a malicious or accidental symlink could cause redacted PDFs to be written outside the intended sandbox. While this is a local tool, the redaction feature is designed to prevent sensitive data leaks, and writing to unexpected locations undermines that guarantee.

## How

1. In `_out_path_is_safe`, resolve the `knowledge_dir` itself using `Path.resolve()` before comparing against the output path.
2. After resolving both paths, verify that the resolved `knowledge_dir` is not a symlink (e.g., check `knowledge_dir.is_symlink()` and reject if true), or alternatively, resolve the output path and confirm it still falls under the resolved `knowledge_dir`.
3. Add a test case that creates a symlink at `knowledge/` pointing to a temporary directory outside the project, then verifies that `_out_path_is_safe` returns `False` for any output path under that symlink.

## Files Affected

- `scripts/pdf_redact.py` (or wherever `_out_path_is_safe` is defined — confirm exact path from repo)
- `tests/test_pdf_redact.py` (or the test file covering `_out_path_is_safe`)

## Constraints

- Must not break existing path validation for normal (non-symlink) cases.
- Must not introduce new dependencies.
- Must not change the behavior of `--check` mode or redaction logic.
- Out of scope: handling symlinks in the input PDF path, or broader filesystem hardening.

## LLM tier

**Sonnet** — This requires moderate design judgment: deciding how to handle symlinks (reject vs. resolve-and-validate), updating the function signature if needed, and adding a test that exercises symlink behavior without being flaky on different platforms.

## Success looks like

- `_out_path_is_safe` returns `False` when `knowledge/` is a symlink to a directory outside the project.
- Existing tests for path safety still pass.
- New test covers the symlink case and passes on Linux/macOS (and Windows if symlink creation is supported).
- No regressions in redaction or `--check` mode.

## Failure looks like

- The function still allows writes through a symlinked `knowledge/` directory.
- The fix breaks normal path validation (e.g., legitimate writes to `knowledge/` are rejected).
- Tests are flaky or fail on platforms where symlink creation requires elevated privileges.

_Follow-up from AI review of PR #145._

---

- **Exact file path for `_out_path_is_safe`**: The function is defined at `projects/08-linkedin-avatar/build/build_profile.py:160` (`def _out_path_is_safe(out_path):`), and called at `projects/08-linkedin-avatar/build/build_profile.py:199`. There is no `scripts/pdf_redact.py`; the actual file is `projects/08-linkedin-avatar/build/build_profile.py`.
- **Exact test file path**: The test file covering this behavior is `projects/08-linkedin-avatar/tests/test_build_profile.py` (contains `test_out_path_outside_knowledge_dir_is_rejected` at line 199). There is no `tests/test_pdf_redact.py`.
- **Current implementation of `_out_path_is_safe`**: The evidence shows the relevant lines in `projects/08-linkedin-avatar/build/build_profile.py`: line 162 `knowledge_dir = KNOWLEDGE_DIR.resolve()` and line 163 `return resolved == knowledge_dir or knowledge_dir in resolved.parents`. The signature is `_out_path_is_safe(out_path)` (line 160), and `knowledge_dir` is not passed in as a parameter — it is derived internally from `KNOWLEDGE_DIR.resolve()`.

## Implementation notes

FILES: projects/08-linkedin-avatar/build/build_profile.py, projects/08-linkedin-avatar/tests/test_build_profile.py
DONE: `_out_path_is_safe` resolves `KNOWLEDGE_DIR` and rejects output paths when `knowledge/` is a symlink pointing outside the project, with a new test covering the symlink case and all existing path-safety tests still passing.

leonarduk:
## What
Updated `_out_path_is_safe` in `projects/08-linkedin-avatar/build/build_profile.py` to reject output paths when `KNOWLEDGE_DIR` is itself a symlink, and added a `TestOutPathIsSafe` test class covering normal, traversal, and symlink cases.

## Why
The previous check resolved the output path but trusted `knowledge/` to be a real directory. If `knowledge/` were replaced with a symlink pointing outside the project, the resolved-containment check could be bypassed and redacted output could be written to an unintended location — undermining the sandbox guarantee the redaction feature is meant to provide.

## Approach
Chose to reject outright when `KNOWLEDGE_DIR.is_symlink()` is true, rather than resolving the symlink and validating containment against its target. A symlinked `knowledge/` is not the layout this tool expects, and even a benign target could be swapped later; refusing to write is the safer, simpler contract. The existing resolve-and-contain check is preserved for the normal case, so non-symlink behavior is unchanged. Alternatives considered: resolving `knowledge/` and comparing against the resolved target (rejected as it silently accepts an unexpected layout), and rejecting only when the resolved target falls outside the project root (rejected as more complex and still permissive).

## Testing
Added `TestOutPathIsSafe` with cases for: path inside `knowledge/`, `knowledge/` itself, path outside, `..` traversal escape, and a symlinked `knowledge/` pointing to an external temp dir (skipped on platforms without symlink support). Existing CLI tests for path rejection and default output continue to exercise the function end-to-end.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #148


leonarduk:
Merged `main`. The security fix itself is correct; found one genuine bug in the PR's own test, unrelated to the merge. `76b5ce7`.

**The core fix is sound and I verified the actual vulnerability, not just the description.** The old check was `resolved == knowledge_dir or knowledge_dir in resolved.parents`, where both `resolved` and `knowledge_dir` are `.resolve()`d. If `KNOWLEDGE_DIR` itself is a symlink to `/tmp/evil`, then `knowledge_dir.resolve()` → `/tmp/evil`, and `out_path = KNOWLEDGE_DIR / "profile.md"` also resolves through that same symlink to `/tmp/evil/profile.md` — so the "containment" check becomes a tautology: both sides pass through the identical symlink and the check can never distinguish a legitimately-nested path from one that escaped via a swapped top-level directory. The new `KNOWLEDGE_DIR.is_symlink()` guard, checked *before* any resolution, closes exactly that gap and nothing else — I confirmed a symlink placed *inside* `knowledge/` (rather than at `knowledge/` itself) is already handled correctly by the pre-existing `.resolve()` call, since it follows every path component, so this PR's narrower scope (reject only when the top-level dir itself is a symlink) is the right fix, not an incomplete one.

**Found a genuine bug in the PR's own branch, independent of the merge.** `test_out_path_outside_knowledge_dir_is_rejected`'s body was:

```python
with pytest.raises(SystemExit):
    build_profile
```

No call — `build_profile` alone is just a module reference, so the exception it asserts on can never fire, and the test failed outright once the merge brought in an unrelated test from `main` immediately after it (`test_default_out_path_is_accepted`, which had gone missing from the merged file for the same reason). Checked directly against the PR's own original commit (`git show bc818ce...`) to confirm this predates any merge conflict — it's truncated in the branch as submitted, not something my merge introduced. `main` already carries the complete, correct version of both tests from an earlier, unrelated PR; restored both from there.

**Testing.** Merged `main`'s own two `TestOutPathIsSafe` additions (`test_parent_of_knowledge_dir_is_rejected`, `test_path_inside_knowledge_dir_is_accepted`) alongside this PR's five — no name collisions, genuinely complementary coverage. Mutation-tested the fix itself: removing the `is_symlink()` guard, inverting it, and dropping the containment check entirely are all caught by the combined suite. 194 passed (189 baseline + 5 new).

flake8 clean. A pre-existing black formatting issue at `build_profile.py:100–104` predates this PR (confirmed directly against `origin/main`) and sits outside its diff, so left untouched.

---

PR #351, opened 2026-09-16
[Issue #157] Wrap `tools.dispatch` in try/except to prevent unhandled exceptions from reaching the UI

### LINKED ISSUE #157: Wrap `tools.dispatch` in try/except to prevent unhandled exceptions from reaching the UI
# Wrap `tools.dispatch` in try/except to prevent unhandled exceptions from reaching the UI

## What

In `_dispatch_tool_call` (referenced at line 111 of the PR), the current try/except only wraps `json.loads`. If `tools.dispatch` itself raises an exception (e.g., a tool implementation error, network failure, or unexpected runtime error), that exception propagates out of `send_message` and reaches the caller as an unhandled stack trace.

The fix: extend the existing exception handling to also catch exceptions raised by `tools.dispatch` and convert them into a friendly error message, consistent with the existing error-handling pattern in the method.

## Why

The acceptance criteria for PR #156 explicitly require "friendly message rather than stack trace" for error cases. The current implementation only handles JSON parse errors from the dispatch result, leaving a gap where a dispatch failure would violate this requirement. This is a correctness risk because tool implementations are external code that can fail in unpredictable ways, and the UI should never surface raw exceptions.

## How

1. In `_dispatch_tool_call`, wrap the `tools.dispatch` call in a try/except block (either extending the existing try or adding a new one) that catches generic `Exception`.
2. In the except handler, log the exception details (for debugging) and return a structured error response that `send_message` can convert into a friendly message, matching the existing error-handling pattern used for JSON parse errors.
3. Ensure the error response includes enough context (e.g., tool name, error message) for the user to understand what failed without exposing internal stack traces.

## Files Affected

- `src/` (the module containing `_dispatch_tool_call` and `send_message` — exact path not confirmed from review text)

## Constraints

- Must not change the existing behavior for successful dispatch calls or JSON parse errors.
- Must not introduce new dependencies.
- Must not alter the public API of `send_message` (signature and return types stay the same).
- Out of scope: retry logic for rate-limit errors, empty-content warnings, and the decorator refactor — these are separate follow-ups.

## LLM tier

**Sonnet** — This requires moderate design judgment to integrate the new exception handling cleanly with the existing error-handling chain without breaking current behavior, and to decide the appropriate error response format.

## Success looks like

- A test exists that simulates `tools.dispatch` raising an exception and verifies `send_message` returns a friendly error message (no stack trace).
- All existing tests in `TestSendMessageHappyPath`, `TestUsageAccounting`, and `TestErrorHandling` still pass.
- Code review confirms the exception is caught and converted to a friendly message, with the original exception logged for debugging.

## Failure looks like

- The exception still propagates to the caller (test fails or manual verification shows a stack trace).
- The fix breaks existing error handling for JSON parse errors or successful dispatch calls.
- The error message is not user-friendly (e.g., exposes internal exception details or raw tool output).

_Follow-up from AI review of PR #156._

---

- The exact file path containing `_dispatch_tool_call` and `send_message` is confirmed: both are in `projects/08-linkedin-avatar/avatar/llm.py` (`_dispatch_tool_call` at line 96, `send_message` at line 112).
- The test file containing `TestSendMessageHappyPath` is identified: `projects/08-linkedin-avatar/tests/test_llm.py` (line 62).

## Implementation notes

FILES: projects/08-linkedin-avatar/avatar/llm.py, projects/08-linkedin-avatar/tests/test_llm.py
DONE: `_dispatch_tool_call` in `avatar/llm.py` catches exceptions from `tools.dispatch`, logs them, and returns a structured error response that `send_message` converts into a friendly message; a new test in `tests/test_llm.py` simulates a dispatch failure and asserts no stack trace reaches the caller, with all existing `TestSendMessageHappyPath`, `TestUsageAccounting`, and `TestErrorHandling` tests still passing.

leonarduk:
## What
Wrapped the `tools.dispatch` call inside `_dispatch_tool_call` in `projects/08-linkedin-avatar/avatar/llm.py` with a try/except that catches any `Exception`, logs it, and returns a structured error result. Added a regression test (`test_dispatch_exception_does_not_crash`) that simulates a tool raising and verifies `send_message` still returns a friendly reply.

## Why
Previously, only `json.loads` was guarded. If a tool implementation raised (network failure, bug, unexpected runtime error), the exception propagated out of `send_message` and surfaced as a raw stack trace in the UI, violating the "friendly message rather than stack trace" acceptance criteria from PR #156. Tool implementations are external code that can fail unpredictably, so the UI must never expose their raw exceptions.

## Approach
Extended the existing error-handling pattern rather than introducing a new one: a nested try/except around `tools.dispatch` mirrors the existing JSON-parse handler and returns a structured `{"error": ..., "detail": ...}` dict that flows through the same downstream path. The tool name is hoisted into a local variable so both handlers can reference it consistently. The original exception is logged via `logger.exception` for debugging, while the user-facing message stays generic (`tool '<name>' failed`) with the exception string in `detail` for context. Alternatives considered: a single combined try/except (rejected — would conflate JSON errors with dispatch errors and lose the distinct log messages), and letting the caller handle it (rejected — the issue explicitly requires the fix at this layer, and callers shouldn't need to know about tool internals).

## Testing
Added `test_dispatch_exception_does_not_crash`, which monkeypatches `llm.tools.dispatch` to raise `RuntimeError`, drives `send_message` through a `FakeClient`, and asserts the reply is returned normally and the tool message payload contains an `error` referencing the tool name. Existing suites (`TestSendMessageHappyPath`, `TestUsageAccounting`, `TestErrorHandling`) continue to pass, confirming no regression in the success path or JSON-parse error path.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #157


leonarduk:
Merged `main` and fixed a real secret-leak bug this PR introduced in `projects/08-linkedin-avatar/avatar/llm.py`. The DeepSeek bot's automated review on this PR already flagged the shape of the problem ("`detail: str(exc)` exposes internal exception details") and was right to request changes — I dug further and found it's not just a friendliness concern, it's a concrete vulnerability specific to this codebase.

**`avatar/tools.py`'s own module docstring guarantees tool implementations never raise** ("None of them may raise — a failed notification or an unknown project name must degrade to an error result, never take down the chat turn"). So reaching the new `except Exception` handler around `tools.dispatch` means some tool broke that contract — the exception that arrives there is, by construction, one nobody vetted for safety to expose.

Concretely: `tools.py`'s `_telegram_notify` already has to defend against exactly this. Its own comment reads:

```python
except requests.RequestException as exc:
    # Telegram's URL embeds the bot token (unlike Pushover's, which keeps
    # it in the POST body) — logger.exception would print that URL, so
    # log only the status code, never the exception object itself.
    status = getattr(exc.response, "status_code", "no response")
    logger.error("Telegram notification failed (status=%s): %s", status, title)
```

`requests.RequestException` messages embed the full request URL, and Telegram's URL embeds the bot token. `_telegram_notify` already knows this and deliberately avoids `logger.exception`/`str(exc)` for it. But if that exception ever escaped past `_telegram_notify`'s own catch (a bug, a new exception type, dispatch routing itself throwing), this PR's `_dispatch_tool_call` would have caught it and put `str(exc)` — bot token included — straight into a `"detail"` field on the tool result, which flows back into the LLM's conversation context on the very next loop iteration. That's a live secret potentially reaching the model (and from there, plausibly the end user) in exactly the failure mode this PR was written to guard against.

I checked issue #157's own AC for the tension the DeepSeek review pointed at: "How" step 3 says the error response should include "enough context... for the user to understand what failed without exposing internal stack traces," while "Failure looks like" separately forbids "expos[ing] internal exception details." Given the concrete leak vector above, I read that as: no exception text of any kind, full stop — the tool name is enough context, and the actual exception is what `logger.exception` (server-side only) is for.

**Fix**: dropped the `"detail": str(exc)` field entirely; `_dispatch_tool_call` now returns just `{"error": f"tool '{tool_name}' failed"}"`, with the real exception still logged server-side via `logger.exception` (unchanged from the PR's own approach there). Added `test_dispatch_exception_message_is_not_leaked_to_llm`, which raises an exception carrying a marker string and asserts it never appears in the tool message content sent back to the LLM — confirmed this test fails against the original `detail: str(exc)` code (reverted locally, re-ran, restored) and passes with the fix. Full `projects/08-linkedin-avatar` suite (191 tests) passes; flake8 clean on the diff (one pre-existing long-line warning on an unrelated line, unchanged by this PR); black clean.

---

PR #372, opened 2026-09-16
[Issue #178] Add automated image size regression checks to CI to prevent large image files from being committed in the future

### LINKED ISSUE #178: Add automated image size regression checks to CI to prevent large image files from being committed in the future
## Summary

Add an automated CI check that fails when newly added or modified image files exceed a size threshold, preventing large image files from being committed in the future.

## What

Add a CI workflow step (or a script invoked by CI) that scans the diff for image files (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, `.svg`) and fails if any added or modified image exceeds a configurable size limit (e.g., 500 KB). The check should run on pull requests and pushes to the default branch.

## Why

PR #177 was an image optimization change via ImgBot that reduced the size of two PNG files. Without an automated guard, large image files can be accidentally committed in future PRs, bloating the repository and degrading performance for anyone cloning or fetching it. This is a maintainability and repository hygiene concern.

## How

1. Create a script (e.g., `scripts/check-image-sizes.sh`) that:
   - Uses `git diff --name-only --diff-filter=AM` against the base branch to get added/modified files.
   - Filters for image extensions.
   - Uses `git cat-file -s` or `stat` to get the byte size of each file.
   - Fails (non-zero exit) if any file exceeds the threshold (e.g., 500 KB), printing the offending file paths and sizes.
2. Add a GitHub Actions workflow (e.g., `.github/workflows/image-size-check.yml`) that:
   - Triggers on `pull_request` and `push` to the default branch.
   - Checks out the code with `fetch-depth: 0` to access the diff.
   - Runs the script with the appropriate base ref (e.g., `origin/main` for PRs).
3. Make the threshold configurable via an environment variable or a constant at the top of the script.

## Files Affected

- `.github/workflows/image-size-check.yml` (new)
- `scripts/check-image-sizes.sh` (new)

## Constraints

- Must not block PRs that only touch non-image files.
- Must not fail on pre-existing large images that are not modified in the PR (only added/modified files are checked).
- Must not interfere with existing CI workflows or deployment processes.
- Out of scope: automatically optimizing images, changing existing image files, or enforcing a specific image format.

## LLM Tier

**Sonnet** — This requires moderate design judgment: writing a shell script with git diff logic, setting up a new CI workflow, and deciding on sensible defaults (threshold, triggers). It is a multi-file change with some heuristic decisions but no architectural ambiguity.

## Success Looks Like

- A PR that adds a new image file larger than 500 KB fails CI with a clear error message listing the file and its size.
- A PR that modifies an existing image to be larger than 500 KB fails CI.
- A PR that only changes code or small images passes CI without the new check interfering.
- The check runs automatically on every PR and push to the default branch.

## Failure Looks Like

- The check passes when a large image is added (false negative) — e.g., due to incorrect diff filtering, wrong base ref, or missing file extensions.
- The check fails on unrelated PRs (false positive) — e.g., it flags pre-existing large images that weren't modified, or it breaks due to missing dependencies or incorrect git context.
- The workflow is not triggered at all due to misconfigured event triggers or path filters.

_Follow-up from AI review of PR #177._

## Implementation notes

FILES: scripts/check-image-sizes.sh, .github/workflows/image-size-check.yml
DONE: A new shell script scans added/modified image files in a PR/push diff and fails with a clear message listing any file exceeding the configurable threshold (default 500 KB), invoked by a new GitHub Actions workflow triggered on `pull_request` and `push` to the default branch with `fetch-depth: 0`, without affecting non-image changes or pre-existing large images.

leonarduk:
## What
Adds a new CI check that fails when newly added or modified image files (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp`, `.svg`) exceed a configurable size threshold (default 500 KB). Includes:
- `scripts/check-image-sizes.sh` — diffs against a base ref, filters added/modified image files, and fails if any exceed the limit.
- `.github/workflows/image-size-check.yml` — runs the script on `pull_request` and pushes to `main`/`master`.

## Why
PR #177 (an ImgBot optimization) highlighted that large image files can be committed without any guardrail, bloating the repository and slowing clones/fetches. This adds an automated regression check to prevent that going forward.

## Approach
- **Diff-based, not repo-wide**: Only added/modified files (`--diff-filter=AM`) are checked, so pre-existing large images don't cause false positives.
- **Shell script + workflow split**: Keeps the logic testable and runnable locally, with the workflow handling ref resolution and checkout.
- **`fetch-depth: 0`** is required so the base ref is available for diffing; the workflow also explicitly fetches `origin/*` and skips gracefully when `github.event.before` is all-zeros (first push).
- **Configurable threshold** via `MAX_IMAGE_SIZE_KB` env var, defaulting to 500 KB.
- **Fail-loud on missing base ref** (exit 2) rather than silently passing, to avoid false negatives.
- **`-z` / `mapfile`** used to safely handle filenames with spaces or newlines.
- **Blob size via `git cat-file -s HEAD:<file>`** preferred over `stat`, with a portable `stat` fallback.
- Rejected alternatives: a repo-wide scan (would flag untouched legacy images) and a third-party action (adds an external dependency for a simple check).

## Testing
- Verified the script locally against a diff containing an oversized image (fails with file path and size), a small image (passes), and non-image changes (no-op).
- Confirmed the workflow triggers on both `pull_request` and `push` to `main`/`master`, and that the first-push all-zeros case exits cleanly.
- Checked that pre-existing large images not present in the diff are ignored.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [ ] No breaking changes

---

Closes #178


leonarduk:
Merged `main` and empirically tested this against the DeepSeek bot's "REQUEST CHANGES" review, since its main claim was specific enough to actually check.

**The bot's headline claim doesn't hold up.** It said `git cat-file -s "HEAD:${file}"` could silently report `size_bytes=0` for a file "present in the diff but not at HEAD" (e.g. added in one commit, removed in a later commit of the same PR). I reproduced exactly that scenario in a scratch repo — add a 600KB image, commit, then `git rm` it, commit again — and ran `git diff --name-only --diff-filter=AM base_ref...HEAD`: it returns nothing. Three-dot diff compares final tree states, so a file that nets out to "doesn't exist at HEAD" can never appear under an A/M filter in the first place — the code path the bot worried about is unreachable.

**But testing surfaced the real version of what it was gesturing at.** The line:
```bash
mapfile -d '' -t CHANGED_FILES < <(
  git diff --name-only --diff-filter=AM -z "${BASE_REF}...HEAD" || true
)
```
can't fail loud *even without* the `|| true`. I confirmed with a minimal repro that `set -e` does not see failures inside a `< <(...)` process substitution at all — a classic bash gotcha, independent of the explicit `|| true`. So a genuinely broken `git diff` (corrupt repo, whatever) would silently look like "no files changed" and exit 0, contradicting this script's own stated design principle ("Fail-loud on missing base ref... to avoid false negatives").

I first considered switching to plain command substitution (`x="$(git diff ...)"`) so `set -e`/`if !` could see the failure — but confirmed with `x="$(printf 'a\0b\0c')"` that bash silently drops embedded NUL bytes in command substitution (`length=${#x}` came back 3, i.e. "abc"), which would have broken the very NUL-safe filename handling `-z` exists for. So instead I redirect `git diff`'s output to a temp file with a normal `>` redirect (whose exit status `if !` *does* see correctly) and read that file into the array with `mapfile`, preserving full NUL-safety while making a genuine failure surface as `exit 2` instead of a silent pass.

Re-ran all the scenarios the PR's own testing section describes, plus the bot's specific edge case, against the patched script in a scratch git repo: large image added → fail; small image added → pass; pre-existing large image untouched → no-op; file added-then-removed in the same range → no-op (confirming the bot's scenario is a non-issue); missing base ref → exit 2; and a filename with spaces → correctly reported. All match the pre-fix behavior, with the added guarantee that a genuine `git diff` failure now surfaces instead of passing silently.

(The bot's other note — hardcoding `main`/`master` instead of reading `github.event.repository.default_branch` — is moot for this repo specifically, since `main` is in fact its default branch; left as-is.)

---

PR #375, opened 2026-09-16
[Issue #185] Consider extracting the rules block to a separate file (e.g., `rules.md`) for easier maintenance and review of prompt content

### LINKED ISSUE #185: Consider extracting the rules block to a separate file (e.g., `rules.md`) for easier maintenance and review of prompt content
## What

Extract the static rules block currently defined as a constant (referred to as `RULES_BLOCK` in the review) into a standalone Markdown file (e.g., `rules.md`) and load it at runtime when assembling the prompt.

The change affects:
- The constant definition (currently `RULES_BLOCK`) — replace its inline string content with a file read.
- The prompt assembly logic that uses this constant (the same location where `budget_for_github` and `max_tokens` are calculated).
- The existing `TestRulesBlock` test class — update tests to read from the new file instead of the inline constant.

## Why

- **Maintainability**: The rules block is prompt content, not code. Keeping it in a `.py` file mixes content with logic, making diffs harder to review and requiring a code change (and potential redeploy) for any wording tweak.
- **Review clarity**: A dedicated `rules.md` file makes prompt-content changes visible as pure content diffs, reducing the risk of accidental code changes sneaking into prompt edits.
- **Protocol compliance**: The rules define AI identity, honesty, scope, and injection posture. A separate file makes it easier to audit these compliance-critical statements without wading through code.

## How

1. Create a new file `rules.md` at the repo root (or alongside the module that currently defines `RULES_BLOCK`) containing the exact current rules text.
2. In the module where `RULES_BLOCK` is defined, replace the inline string with a file read, e.g.:
   ```python
   from pathlib import Path
   RULES_BLOCK = Path(__file__).parent / "rules.md"
   ```
   (Adjust the path relative to the actual file location.)
3. Update the prompt assembly logic to use the loaded content — no other logic changes should be needed since the constant name stays the same.
4. Update `TestRulesBlock` to load the rules from the file (or keep the constant as the source of truth and add a test that the file content matches the constant).
5. Ensure the file is included in packaging (if the project is distributed) so the runtime read works in all environments.

## Files Affected

- `rules.md` (new file)
- The module containing the `RULES_BLOCK` constant (exact path unknown from review text — likely the prompt-assembly module)
- The test file containing `TestRulesBlock` (exact path unknown from review text)

## Constraints

- **Do not change the rules content** — this is a pure refactor. The exact text, ordering, and placement (last section in the assembled prompt) must remain identical.
- **Do not break the `max_tokens` / `budget_for_github` calculation** — the token budget logic must continue to use the same `RULES_BLOCK` constant name.
- **Do not alter the `test_rules_block_is_the_last_section_in_the_assembled_prompt` or `test_is_static_text_with_no_interpolation_markers` tests** — they must still pass unchanged.
- **Out of scope**: Adding the edge-case test for very small `max_tokens` (that is a separate follow-up), and any changes to the rules content itself.

## LLM tier

**Sonnet** — This is a multi-file refactor (new file, constant change, test updates) that requires moderate design judgment about file placement and how to load the content, but it is not architecturally complex or ambiguous.

## Success looks like

- `rules.md` exists and contains the exact current rules text.
- The `RULES_BLOCK` constant is loaded from the file at runtime.
- All existing tests pass, including `TestRulesBlock` (with tests updated to read from the file if needed).
- The assembled prompt is byte-for-byte identical to before the change.
- No changes to the rules content, token budget logic, or prompt placement.

## Failure looks like

- The rules text differs from the original (accidental edit during extraction).
- The prompt assembly breaks because the file path is wrong or the file is not packaged.
- Tests fail because they still reference the inline constant that no longer exists.
- The rules block is no longer the last section in the assembled prompt.
- The `max_tokens` calculation changes behavior, causing token budget regressions.

_Follow-up from AI review of PR #183._

---

- Which module/file currently defines `RULES_BLOCK`? — `projects/08-linkedin-avatar/avatar/context.py` (defined at line 35, used at lines 173 and 185).
- Which test file contains `TestRulesBlock`? — `projects/08-linkedin-avatar/tests/test_context.py` (class `TestRulesBlock` at line 203).

## Implementation notes

FILES: projects/08-linkedin-avatar/avatar/rules.md, projects/08-linkedin-avatar/avatar/context.py, projects/08-linkedin-avatar/tests/test_context.py
DONE: `rules.md` exists alongside `context.py` containing the exact current rules text, `RULES_BLOCK` is loaded from that file at runtime, and all existing tests (including `TestRulesBlock`) pass with the assembled prompt byte-for-byte identical to before.

leonarduk:
## What
Extracts the static `RULES_BLOCK` prompt content out of `avatar/context.py` into a standalone `avatar/rules.md` file, and loads it at import time via `Path(__file__).resolve().parent / "rules.md"`. The `RULES_BLOCK` constant name and value are unchanged, so prompt assembly, token budgeting, and placement logic are untouched. Adds a test asserting the loaded constant matches the file contents.

## Why
The rules block is prompt content, not code. Keeping it inline in a `.py` file mixes content with logic, makes wording changes show up as code diffs, and requires a code change (and potential redeploy) for any copy tweak. A dedicated `rules.md` makes prompt edits reviewable as pure content diffs and makes the compliance-critical statements (AI identity, honesty, scope, injection posture) easier to audit.

## Approach
Load the file once at import time rather than per call: the text is static, so re-reading per request buys nothing, and a missing file fails loudly at startup instead of silently shipping a prompt with no rules. Kept the `RULES_BLOCK` constant name so the token budget (`max_tokens` / `budget_for_github`) and prompt assembly code need no changes. Placed `rules.md` alongside the module (`avatar/rules.md`) so the relative path is stable regardless of working directory. Rejected keeping the inline string as source of truth with a sync test — that would leave the content in code and defeat the purpose.

## Testing
- Added `test_rules_block_is_loaded_from_rules_md`, which asserts `avatar/rules.md` exists and that `context.RULES_BLOCK` equals its stripped contents — covering both the file and the loader.
- Existing `TestRulesBlock` assertions run unchanged against the loaded content.
- Verified the assembled prompt is byte-for-byte identical to before the refactor (same text, same ordering, rules still last section).

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #185


leonarduk:
Merged `main` and checked both of the DeepSeek bot's "REQUEST CHANGES" claims empirically.

**The extraction itself is correct.** I reconstructed the original inline `RULES_BLOCK` Python string concatenation and diffed it byte-for-byte against `avatar/rules.md`'s content (after the same `.strip()` the loader applies) — they match exactly. So the "byte-for-byte identical" acceptance criterion genuinely holds for this PR as submitted.

**But the bot's underlying test-coverage point was right**: the PR's only new test (`test_rules_block_is_loaded_from_rules_md`) asserts `RULES_BLOCK == rules_path.read_text()...` — that's tautological with respect to the AC. It would still pass even if `rules.md`'s content had drifted from the original during extraction (a re-wrapped line, a dropped word), since it never compares against the pre-refactor text, only the loader against the file it loads. Added `test_rules_block_content_matches_exact_expected_text`, which pins the exact original string verbatim and asserts `RULES_BLOCK` still equals it — this is the test that actually enforces the AC. Confirmed it works: temporarily edited `rules.md`'s content, watched this new test fail with a clear diff, then restored it and confirmed the full suite (191 tests) is green again.

**The bot's other blocking claim — packaging — doesn't apply to this project.** It flagged that `rules.md` might not ship in an installed wheel without `package_data`/`MANIFEST.in` config. I checked: there's no `pyproject.toml` or `setup.py` anywhere in this project, and per `docs/deployment.md`, Render deploys it via `pip install -r requirements.txt` directly against a git checkout — the app is never packaged as a wheel. `rules.md` already sits on disk next to `context.py` in that checkout, exactly where `Path(__file__).resolve().parent / "rules.md"` expects it, so there's nothing to fix here.

flake8 and black clean on the diff (using this repo's actual CI thresholds — `--select=E9,F63,F7,F82` for the blocking check and `--max-line-length=127` for the advisory one, per `.github/workflows/python-ci.yml`, rather than a stricter default).

---

PR #419, opened 2026-09-16
[Issue #215] Add a self-test (or a `workflow_dispatch` dry-run) that verifies the mcp-pin check fails when the pin is removed, to guard against the check silently becoming a no-op.

### LINKED ISSUE #215: Add a self-test (or a `workflow_dispatch` dry-run) that verifies the mcp-pin check fails when the pin is removed, to guard against the check silently becoming a no-op.
## What

Add a self-test (or a `workflow_dispatch` dry-run) that verifies the mcp-pin CI check in `.github/workflows/python-ci.yml` actually fails when the pin is removed. The check currently iterates over `projects/01-mcp-server-suite/servers/*/requirements.txt`, greps for `mcp<2.0.0`, and exits non-zero on any miss. There is no test proving the check is not a silent no-op.

## Why

The mcp-pin check is the only guard enforcing `mcp<2.0.0` across server requirements files. If the glob stops matching (e.g., directory layout changes, files renamed, `shopt -s nullglob` guard fires on zero matches), the check passes vacuously and provides false confidence. A negative test ensures the check fails loudly when the pin is absent, so the guard cannot silently degrade into a no-op.

## How

Add a self-test that exercises the check against a known-bad input:

- Preferred: a `workflow_dispatch`-triggered job (or a step gated on `workflow_dispatch`) that creates a temporary fixture `requirements.txt` without the pin, runs the same grep/loop logic against it, and asserts a non-zero exit. Fail the job if the check unexpectedly passes.
- Alternative: extract the check logic into a small shell script (e.g., under `.github/scripts/` or similar) invoked by both the CI step and the self-test, so the test exercises the real code path rather than a copy.
- Keep the self-test out of the default PR path if it would add noise; `workflow_dispatch` is acceptable per the issue title.

## Files Affected

- `.github/workflows/python-ci.yml`
- Unknown (only if extracting the check into a reusable script)

## Constraints

- Must not change the existing mcp-pin check's behavior on real PRs (still runs after "Verify shared repository automation" and before flake8).
- Must not modify server logic, requirements content, or unrelated CI steps.
- Must not introduce new Python dependencies or `pyproject.toml` changes.
- Out of scope: anchoring the grep to `^mcp<2\.0\.0`, documenting the pin in the project README, and any non-blocking mechanism for PR #28 (already merged).

## LLM tier

Sonnet — requires design judgment on where to place the self-test and whether to extract shared logic, but is scoped to a single workflow file.

## Success looks like

- A `workflow_dispatch` (or equivalent) run exists that deliberately omits the pin and observes the check exit non-zero with the `::error::FAIL: ...` message.
- The self-test fails the job if the check unexpectedly passes on the bad fixture.
- The existing PR-triggered mcp-pin step is unchanged in behavior and ordering.
- No changes to server code, requirements files, or unrelated workflow steps.

## Failure looks like

- The self-test passes regardless of whether the pin is present (i.e., it does not actually exercise the check).
- The self-test duplicates the grep logic instead of invoking the real check, so drift between the two is possible.
- The existing mcp-pin step is altered, reordered, or made conditional in a way that weakens enforcement on normal PRs.
- New dependencies, permissions, or unrelated CI changes are introduced.

_Follow-up from AI review of PR #214._

## Implementation notes

FILES: .github/workflows/python-ci.yml, .github/scripts/check_mcp_pin.sh
DONE: A `workflow_dispatch`-triggered self-test (or a step gated on `workflow_dispatch`) runs the mcp-pin check against a temporary fixture `requirements.txt` lacking the pin and fails the job if the check exits zero; the existing PR-triggered mcp-pin step remains unchanged in behavior and ordering (after "Verify shared repository automation", before flake8).

leonarduk:
## What
Extracts the mcp-pin check from `.github/workflows/python-ci.yml` into a reusable script at `.github/scripts/check_mcp_pin.sh`, wires the CI step to invoke it, and adds a `workflow_dispatch`-gated `mcp-pin-selftest` job that exercises the script against known-bad fixtures.

## Why
The mcp-pin check is the only guard enforcing `mcp<2.0.0` across MCP server requirements files. If the glob stops matching (directory renames, layout changes, zero matches) or the grep silently stops matching, the check passes vacuously and provides false confidence. There was no test proving the check isn't a silent no-op. This change adds a negative test so the guard fails loudly when the pin is absent or when no requirements files are discovered.

## Approach
Extracted the check into `.github/scripts/check_mcp_pin.sh` (the issue's preferred alternative) so the self-test invokes the real code path rather than duplicating the grep/loop logic — avoiding drift between the CI step and the test. The script also treats zero discovered requirements files as a failure, closing the vacuous-pass hole directly in the check itself. The self-test job is gated on `github.event_name == 'workflow_dispatch'` to keep it out of the default PR path, per the issue. The existing PR-triggered step is preserved in place (after "Verify shared repository automation", before flake8) and now simply calls the script, so behavior and ordering on real PRs are unchanged.

## Testing
The `mcp-pin-selftest` job (run via `workflow_dispatch`) performs three verifications:
1. Runs the script against the real repository and asserts it passes.
2. Builds a temp fixture with a `requirements.txt` missing the pin, runs the script, and asserts a non-zero exit **and** that the output contains the expected `::error::FAIL:` message.
3. Runs the script against an empty fixture root and asserts a non-zero exit, confirming the check does not pass vacuously when no requirements files are found.

Each step fails the job with a distinct `::error::` message if the check unexpectedly passes.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #215


leonarduk:
Merged `main` (a real conflict — `main` had independently added a `pricing-staleness` job and a `schedule:` trigger to the same file since this branch was cut) and fixed the exact bug the DeepSeek bot found, plus its root cause.

**DeepSeek's specific claim checked out.** I reproduced it directly: the self-test's negative fixture uses the comment `# intentionally missing the mcp<2.0.0 pin`, which itself contains the literal substring `mcp<2.0.0`. `grep -q 'mcp<2\.0\.0'` (the extracted script's pattern) matches that comment line, so `check_mcp_pin.sh` reports the fixture as passing — meaning the self-test's own assertion ("the check must fail here") would fire and the `mcp-pin-selftest` job would always fail on `workflow_dispatch`, for the wrong reason.

**But the merge surfaced something better than a wording fix.** `main` had independently hardened the *inline* version of this same check (before this PR extracted it into a script) to use an anchored regex — `grep -qE '^mcp.*<2\.0\.0'` instead of the unanchored `grep -q 'mcp<2\.0\.0'` this PR's extracted script uses. Anchoring to the start of the line means a comment merely *mentioning* the pin string can never satisfy the check — only an actual requirement line can. That's a strictly better fix than just rewording the fixture's comment (which would still leave the check itself fragile against the next similarly-worded comment anyone adds). Took `main`'s anchored pattern into the extracted `check_mcp_pin.sh`, and rewrote the fixture's comment anyway for clarity (no longer contains the pin string at all, so it stays a valid negative case regardless).

Verified all three self-test scenarios directly against the merged, fixed script:
- Real repo → passes (exit 0)
- Fixture missing the pin (fixed comment wording) → fails with the expected `::error::FAIL:` message (exit 1)
- Empty fixture root → fails as a vacuous-pass guard (exit 1)
- Bonus: a fixture with a comment *mentioning* the pin *and* an actual pin line on its own → still correctly passes (confirms the anchoring doesn't create a false negative)

`actionlint` and YAML parsing both clean on the merged workflow file.

---

PR #426, opened 2026-09-16
[Issue #220] Align non-interactive `tokens_per_sec` lower bound with the interactive prompt's `minimum=0.001` (or document why they differ).

### LINKED ISSUE #220: Align non-interactive `tokens_per_sec` lower bound with the interactive prompt's `minimum=0.001` (or document why they differ).
## What

The non-interactive config validation for `tokens_per_sec` in `run_non_interactive()` currently rejects only values `<= 0`, while the interactive prompt enforces `minimum=0.001`. This issue tracks either:

1. Tightening the non-interactive guard to reject values `< 0.001` (matching the interactive prompt), **or**
2. Documenting in code (comment/docstring) and/or user-facing docs why the two modes intentionally differ.

## Why

- **Consistency / least surprise:** Users switching between interactive and non-interactive modes should not see different acceptance thresholds for the same field. A value like `0.0001` is accepted non-interactively but rejected interactively.
- **Correctness risk is low** (the non-interactive bound is more permissive, not incorrect), but the divergence is a latent source of confusion and future bug reports.
- The original issue text explicitly said the guard should "mirror the interactive prompt's `minimum` check," so the current implementation does not fully satisfy the stated intent.

## How

1. Locate the `tokens_per_sec` validation inside `run_non_interactive()` (the same block that now includes the `isinstance(tokens_per_sec, bool)` check and raises `ConfigError`).
2. Decide on one of:
   - **Align:** change the lower-bound check from `tokens_per_sec <= 0` to `tokens_per_sec < 0.001`, and update the `ConfigError` message to state the minimum (`0.001`) rather than just "must be positive."
   - **Document:** keep `> 0` and add a short comment explaining the intentional divergence (e.g., non-interactive callers may legitimately want very small rates), plus a note in the relevant docs.
3. If aligning, update the existing parametrized test and `test_run_non_interactive_rejects_zero_tokens_per_sec_in_every_mode` so the boundary cases reflect the new threshold (e.g., add `0.0001` as a rejected value, `0.001` as accepted).
4. If documenting, add a test asserting `0.0001` is accepted, to lock in the intentional behavior.

## Files Affected

- Unknown (the module containing `run_non_interactive()` and its `tokens_per_sec` validation)
- Unknown (the test module containing `test_run_non_interactive_rejects_zero_tokens_per_sec_in_every_mode` and the parametrized `tokens_per_sec` test)

## Constraints

- Must not change the CLI signature or introduce new dependencies.
- Must continue raising `ConfigError` (existing callers and tests depend on this type).
- Must preserve the `isinstance(tokens_per_sec, bool)` rejection — `True` must remain invalid.
- Out of scope: changing the interactive prompt's `minimum=0.001`, and any refactor of mode-specific branching.

## LLM tier

**Haiku** if the decision is to align (mechanical threshold + message + test tweak). **Sonnet** if the decision is to document the divergence, since it requires a judgment call on wording and where to surface the rationale.

## Success looks like

- Either the non-interactive guard rejects `< 0.001` with a `ConfigError` message naming the field and the minimum, **or** the divergence is explicitly documented in code and a test asserts `0.0001` is accepted.
- Tests cover the boundary: `0.001` accepted, `0.0009` (or `0.0001`) rejected if aligning; `0.0001` accepted if documenting.
- All three modes (`own`, `existing`, `rent`) remain covered by the rejection test.
- No change to `ConfigError` type or CLI surface.

## Failure looks like

- The threshold is changed but the error message still says "must be positive," leaving the actual bound unclear.
- The `bool` rejection is accidentally dropped or reordered such that `True` passes.
- Only one mode is tested for the new boundary.
- The interactive prompt's `minimum=0.001` is changed as a side effect.

_Follow-up from AI review of PR #219._

---

- `run_non_interactive()` is defined in `projects/05-llm-cost-comparison/llm_cost_comparison.py:2197`; its `tokens_per_sec` validation lives in the same file at lines 426–427 (`if tokens_per_sec <= 0: raise ValueError("tokens_per_sec must be > 0")`), inside the function taking `total_monthly_tokens, tokens_per_sec` at line 417.
- The test module is `projects/05-llm-cost-comparison/test_llm_cost_comparison.py`; `test_run_non_interactive_rejects_zero_tokens_per_sec_in_every_mode` is at line 1253, and parametrized `tokens_per_sec` usages appear at lines 63, 81, 94, 160, 182.
- The three modes referenced (`own`, `existing`, `rent`) map to code paths in `projects/05-llm-cost-comparison/llm_cost_comparison.py` and `test_llm_cost_comparison.py`; specifically `rent` mode is exercised by `test_run_non_interactive_rent_mode` (line 1012) and `test_run_non_interactive_rent_mode_missing_hourly_rate_raises_config_error` (line 1033), and `run_non_interactive` is invoked from the CLI at line 2437.

## Implementation notes

DONE: The non-interactive `tokens_per_sec` guard either rejects values `< 0.001` with a `ConfigError` message naming the field and minimum (aligning with the interactive prompt), or the intentional divergence is documented in code with a test asserting `0.0001` is accepted; boundary tests cover `0.001` accepted and `0.0009`/`0.0001` rejected (or `0.0001` accepted if documenting) across all three modes (`own`, `existing`, `rent`), while preserving the `bool` rejection and `ConfigError` type.

leonarduk:
## What
Tightened the non-interactive `tokens_per_sec` validation in `run_non_interactive()` to reject values `< 0.001` (previously `<= 0`), matching the interactive prompt's `minimum=0.001` guard. Updated the `ConfigError` message to name the minimum, and expanded tests to cover the new boundary.

## Why
The interactive prompt rejected `tokens_per_sec < 0.001`, but the non-interactive path accepted any positive value (e.g. `0.0001`). This divergence was a latent source of confusion for users switching between modes and did not satisfy the original intent that the non-interactive guard "mirror the interactive prompt's `minimum` check."

## Approach
Chose to **align** the thresholds rather than document the divergence. Aligning is the least-surprise option and matches the issue's stated intent; documenting would have required justifying why the same field has different acceptance rules across modes, which offers no clear user benefit. The `isinstance(..., bool)` rejection is preserved and ordered before the numeric comparison so `True` remains invalid. The interactive prompt's `minimum=0.001` was left untouched, and no CLI surface or `ConfigError` type changed.

## Testing
- Extended the parametrized rejection test with `0.0009` and `0.0001` as bad values.
- Renamed and re-parametrized the per-mode rejection test to cover `0`, `0.0009`, and `0.0001` across all three modes (`own`, `existing`, `rent`).
- Added a new per-mode acceptance test asserting `0.001` (the inclusive boundary) succeeds in every mode.
- Ran the full test module; all tests pass.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #220


leonarduk:
Merged `main` and verified this independently — DeepSeek's APPROVE verdict holds up. Ran the full 173-test suite in `projects/05-llm-cost-comparison` (all pass) and confirmed the interactive prompt's `minimum=0.001` literal genuinely matches the new non-interactive threshold, so the two paths are actually aligned now, not just documented as aligned.

Applied the one accurate, cheap nit from the review: renamed `test_run_non_interactive_rejects_nonpositive_tokens_per_sec` to `test_run_non_interactive_rejects_invalid_tokens_per_sec`, since the parametrize list now also covers sub-minimum *positive* values (`0.0009`, `0.0001`) that "nonpositive" doesn't describe. Left the other nit (the new acceptance test asserting `"Claude Opus 5"` appears in output) as-is per the review's own "non-blocking" — it's a stable fixture value in this test file already used elsewhere, not a real coupling risk.

flake8 and black both clean on the diff.

---

PR #432, opened 2026-09-16
[Issue #225] Verify `interactive_local_setup()` is not reachable in a context where the new unconditional prompt would fire unexpectedly.

### LINKED ISSUE #225: Verify `interactive_local_setup()` is not reachable in a context where the new unconditional prompt would fire unexpectedly.
## What

Verify that `interactive_local_setup()` in `projects/05-llm-cost-comparison/llm_cost_comparison.py` is only ever invoked after the user has already opted into local setup, so that the newly added unconditional combined prompt:

```
Skip benchmark (GPU detection + throughput)? [y/N]:
```

cannot fire in a context where the user did not intend to configure local setup.

If the function *is* reachable without a prior opt-in gate, either (a) move the combined prompt below the existing opt-in gate, or (b) add an explicit guard so the prompt only fires once local setup has been confirmed.

## Why

PR #222 added the combined skip-benchmark prompt at the top of `interactive_local_setup()`, before any of the pre-existing context in that function. The review noted that if the original flow had any early-return or short-circuit before the GPU prompt (e.g. a "do you want to configure local setup at all?" gate), the new prompt would now fire even when the user did not want local setup. The reviewer could not confirm from the diff alone whether this is a real bug, so it was filed as a non-blocking follow-up. This issue tracks that verification and any resulting fix.

## How

1. Read the full body of `interactive_local_setup()` in `projects/05-llm-cost-comparison/llm_cost_comparison.py`, including any lines above the new combined prompt.
2. Trace every call site of `interactive_local_setup()` and confirm whether each one is already gated behind a user opt-in to local setup (e.g. a prior `prompt_yes_no` or CLI flag check).
3. If every call site is gated: add a short comment above the combined prompt documenting the invariant (e.g. "callers must have already confirmed local setup"), and close the issue with a note.
4. If any call site is *not* gated: move the combined prompt below the existing opt-in gate, or add an explicit early-return/guard so the prompt only fires after opt-in. Preserve the existing `default=False` semantics and the downstream gating on `gpu_detection_enabled` / `benchmark_enabled`.
5. Do not change the prompt text, default, or the two original `prompt_yes_no` calls.

## Files Affected

- `projects/05-llm-cost-comparison/llm_cost_comparison.py`

## Constraints

- Must not change the behavior verified in PR #222: `y` skips both GPU detection and benchmarking; `n`/Enter falls through to the original two-question flow; non-interactive mode is untouched.
- Must not alter the prompt string, its `default=False`, or the existing `prompt_yes_no` calls.
- Out of scope: the docstring/comment updates and regression tests called out as separate follow-ups in the same review.

## LLM tier

Sonnet — requires reading the full function and tracing call sites to decide whether a code change is needed, but the change itself (if any) is small and localized.

## Success looks like

- A written determination in the issue thread stating whether `interactive_local_setup()` is reachable without a prior local-setup opt-in, with the specific call sites cited.
- If reachable: a minimal patch that ensures the combined prompt only fires after opt-in, with the PR #222 behavior preserved.
- If not reachable: an inline comment documenting the invariant, and the issue closed as verified.
- No change to the prompt text, default, or downstream gating logic.

## Failure looks like

- The combined prompt is moved or guarded in a way that changes the `y` / `n` / Enter behavior verified in PR #222.
- The prompt is duplicated or the original two `prompt_yes_no` calls are modified.
- The issue is closed without a concrete statement of which call sites were checked and what was found.

_Follow-up from AI review of PR #222._

## Implementation notes

DONE: A written determination in the issue thread states whether `interactive_local_setup()` is reachable without a prior local-setup opt-in, citing each call site; if reachable, a minimal patch ensures the combined skip-benchmark prompt only fires after opt-in (or an inline comment documents the invariant if all call sites are gated), with PR #222 `y`/`n`/Enter behavior and non-interactive mode preserved.

leonarduk:
## What
Added an inline comment above the combined skip-benchmark prompt in `interactive_local_setup()` documenting the invariant that all callers must have already confirmed the user wants to configure local setup. No functional code changes.

## Why
PR #222 introduced an unconditional combined prompt (`Skip benchmark (GPU detection + throughput)? [y/N]:`) at the top of `interactive_local_setup()`. A reviewer flagged that if any caller reached this function without a prior local-setup opt-in, the prompt would fire unexpectedly. This issue tracked verifying whether that was a real bug and fixing it if so.

## Approach
Traced every call site of `interactive_local_setup()`. The only caller is `run_interactive()`, which invokes it unconditionally as part of the main interactive flow — there is no upstream "skip local setup entirely" gate. Therefore the combined prompt is the first local-setup question the user sees, matching pre-PR #222 behavior where the GPU detection prompt was also first. Since no call site is ungated, no code change was needed; per the issue's step 3, added a comment documenting the invariant so future callers adding an opt-in gate know to move the prompt below it. Rejected alternatives: moving the prompt or adding a guard (unnecessary given no ungated callers) and modifying the prompt text/default (explicitly out of scope).

## Testing
Verified by reading the full body of `interactive_local_setup()` and tracing all call sites via grep. Confirmed the only caller is `run_interactive()`, which has no local-setup opt-in gate upstream. No behavioral change, so PR #222 semantics (`y` skips both steps; `n`/Enter falls through to the two original prompts; non-interactive mode untouched) are preserved by construction.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #225


leonarduk:
Merged `main` and want to walk through why I'm keeping the substance of this PR despite the DeepSeek bot's "REQUEST CHANGES" — it was right about a real writing problem, but wrong about there being a real bug.

**The self-contradiction it caught was real.** The comment said "every caller ... must have already confirmed the user wants to configure local setup" and then, in the very next clause, "there is no ... gate upstream" — those two statements can't both be true, and a future reader genuinely couldn't tell from it whether the invariant holds.

**But its conclusion — that this documents a live bug matching the issue's step 4 ("if reachable without opt-in, add a guard or move the prompt") — doesn't hold up.** I traced the full call chain directly rather than trusting either the PR's or the bot's framing:

- `interactive_local_setup()` has exactly one caller: `run_interactive()`.
- `run_interactive()` is called unconditionally from `main()` whenever `--non-interactive` isn't passed — it *is* the tool's interactive mode.
- `run_interactive()`'s own docstring states its flow always asks for "a workload preset ..., local hardware setup, and hosted model selection" — three fixed, sequential steps, not an optional branch behind a yes/no gate.

This tool's entire purpose is comparing local vs. hosted costs. There's no "skip local setup entirely" concept anywhere in it to opt into or gate — local setup input is always needed, the same way it always was before PR #222 (the original GPU-detection prompt was equally unconditional and first). So there's genuinely nothing to guard against here; the issue's own premise (that an ungated reachability might be a bug) doesn't apply to this codebase's actual design.

Rewrote the comment to state that plainly — no "invariant" framing that implies a violable precondition, no self-contradiction, and it cites the actual traced call chain (`interactive_local_setup()` ← `run_interactive()` ← `main()`) so a future reader has the real trace, not an assertion.

Full suite (162 tests) passes; flake8 and black clean.

---

PR #433, opened 2026-09-16
[Issue #227] Confirm `_resolve_workload_scenarios` always returns object-shaped workloads (not dicts) so `getattr` in the new validation loop is safe.

### LINKED ISSUE #227: Confirm `_resolve_workload_scenarios` always returns object-shaped workloads (not dicts) so `getattr` in the new validation loop is safe.
## What

Confirm that `_resolve_workload_scenarios(config)` always returns scenarios whose workload element is an object exposing `requests_per_day`, `avg_input_tokens`, and `avg_output_tokens` as attributes — never a `dict` or other mapping.

The new validation loop added in PR #226 (inside `run_non_interactive`) iterates over the scenarios returned by `_resolve_workload_scenarios` and calls `getattr(_workload, _field)` for each of the three fields. If any scenario's workload is a `dict`, this raises `AttributeError` instead of the intended `ConfigError`, defeating the purpose of the validation.

## Why

- **Correctness risk:** The validation is meant to raise `ConfigError` with a clear message when a workload field is non-positive. If the workload shape is ever a mapping, users get an opaque `AttributeError` instead — the exact failure mode this validation was added to prevent.
- **Silent contract:** The new loop implicitly depends on the return shape of `_resolve_workload_scenarios`. That contract is not documented or asserted anywhere, so a future refactor of `_resolve_workload_scenarios` could break validation without any test catching it.
- **Non-blocking follow-up:** Flagged during AI review of PR #226 as something to confirm rather than block on, since existing end-to-end tests (`test_run_non_interactive_end_to_end`, `test_run_non_interactive_rent_mode`) pass through this path.

## How

1. Locate `_resolve_workload_scenarios` (in the same module as `run_non_interactive`) and inspect its return type. Determine whether the workload element of each scenario tuple is guaranteed to be an object with the three attributes.
2. If it is guaranteed:
   - Add a short docstring or type annotation on `_resolve_workload_scenarios` documenting the return shape (e.g. `Iterable[tuple[str, str, WorkloadConfig]]` or equivalent).
   - Optionally add a defensive assertion or a narrow `isinstance` check in the validation loop that raises `ConfigError` (not `AttributeError`) if the workload is not object-shaped, so the failure mode stays user-friendly.
3. If it is *not* guaranteed (e.g. some code paths return dicts):
   - Normalize the workload to an object before the validation loop, or
   - Switch the validation loop to read fields via a small helper that handles both attribute and mapping access, or
   - Fix `_resolve_workload_scenarios` to always return objects.
4. Add a regression test that exercises the validation loop against a scenario shape that would previously have triggered `AttributeError`, asserting that `ConfigError` is raised (or that validation passes cleanly).

## Files Affected

- The module containing `run_non_interactive` and `_resolve_workload_scenarios` — Unknown (path not given in review text)
- The test module containing `test_run_non_interactive_end_to_end` and `test_run_non_interactive_rent_mode` — Unknown (path not given in review text)

## Constraints

- Must not change the user-facing behavior of the validation for the existing single-workload config shape: zero or negative `requests_per_day`, `avg_input_tokens`, or `avg_output_tokens` must still raise `ConfigError` with the `workload.<field>` message format introduced in PR #226.
- Must not weaken validation to skip non-primary scenarios; the current loop validates all scenarios and that behavior is intended.
- Out of scope: changing the error message wording, adding multi-scenario validation tests beyond the regression test above, or refactoring `_resolve_workload_scenarios` for unrelated reasons.

## LLM tier

**Sonnet** — requires reading `_resolve_workload_scenarios` to determine the actual return shape, then deciding between documentation-only, defensive check, or normalization. Moderate design judgment, not a mechanical one-liner.

## Success looks like

- The return shape of `_resolve_workload_scenarios` is either documented and asserted, or normalized so the validation loop cannot raise `AttributeError`.
- A regression test exists that would fail if the workload element were a `dict` (i.e. it asserts `ConfigError` or clean pass, not `AttributeError`).
- Existing tests (`test_run_non_interactive_end_to_end`, `test_run_non_interactive_rent_mode`, and the parametrized zero/negative test for the three fields) still pass unchanged.

## Failure looks like

- The validation loop still calls `getattr` on a value whose shape is unverified, with no test guarding against a mapping-shaped workload.
- The fix changes the `ConfigError` message format or the set of scenarios validated, breaking the PR #226 contract.
- A defensive check is added that swallows the error silently instead of

---

- The module containing `run_non_interactive` and `_resolve_workload_scenarios` is `projects/05-llm-cost-comparison/llm_cost_comparison.py` (definitions at lines 2197 and 2137 respectively).
- The test module containing `test_run_non_interactive_end_to_end` and `test_run_non_interactive_rent_mode` is `projects/05-llm-cost-comparison/test_llm_cost_comparison.py` (lines 638 and 1012 respectively).
- The file path for the PR #226 validation loop is `projects/05-llm-cost-comparison/llm_cost_comparison.py`; the loop's call site is at line 1913 (`scenarios = _resolve_workload_scenarios(settings)`), and `_resolve_workload_scenarios` is defined at line 2137 with signature `def _resolve_workload_scenarios(config: dict) -> list:`.

## Implementation notes

DONE: `_resolve_workload_scenarios` documents/asserts its return shape (or normalizes workloads to objects), the validation loop raises `ConfigError` rather than `AttributeError` for mapping-shaped workloads, and a regression test covers that path while existing tests still pass.

leonarduk:
## What
Documented the return-shape contract of `_resolve_workload_scenarios` (third tuple element is always a `Workload` instance, never a mapping) and added a defensive `isinstance` check in `run_non_interactive`'s scenario loop that raises `ConfigError` if a non-`Workload` value ever appears. Added two regression tests covering the contract and the failure mode.

## Why
The validation loop added in PR #226 calls `getattr(_workload, _field)` on each scenario's workload. If `_resolve_workload_scenarios` ever returned a `dict`-shaped workload, users would get an opaque `AttributeError` instead of the intended `ConfigError` — the exact failure mode the validation was meant to prevent. The implicit contract was undocumented and untested, so a future refactor could silently break it.

## Approach
Inspection confirmed both branches of `_resolve_workload_scenarios` construct `Workload` objects explicitly (`Workload(**config["workload"])` and `preset.to_workload()`), so the contract already holds. Rather than normalizing or adding a mapping-aware accessor (unnecessary given the guarantee), I:
- Documented the contract in the docstring so future refactors see it.
- Added a narrow `isinstance(workload, Workload)` guard that raises `ConfigError` with a clear "internal error" message, preserving the user-facing failure mode if the contract is ever violated.
- Left the existing validation loop and `ConfigError` message format untouched.

## Testing
- `test_resolve_workload_scenarios_always_returns_workload_objects` asserts every scenario's third element is a `Workload` with the three required attributes, across explicit-workload, single-preset, and multi-preset configs.
- `test_run_non_interactive_raises_config_error_for_mapping_shaped_workload` monkeypatches the resolver to return a `dict`-shaped workload and asserts `run_non_interactive` raises `ConfigError` matching `"not a Workload"` (not `AttributeError`).
- Existing tests (`test_run_non_interactive_end_to_end`, `test_run_non_interactive_rent_mode`, and the parametrized zero/negative field tests) pass unchanged.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed
- [x] No breaking changes

---

Closes #227


leonarduk:
Merged `main` (clean, no conflicts) and verified this beyond just trusting the description — DeepSeek's APPROVE holds up.

Ran the full 164-test suite (all pass), then mutation-tested the actual defensive check: temporarily removed the `isinstance(workload, Workload)` guard and re-ran the regression test. Without the guard, it fails with exactly the opaque `AttributeError: 'dict' object has no attribute 'monthly_total_tokens'` this PR was written to prevent; with the guard restored, it raises the clean `ConfigError` instead. So the test genuinely exercises the fix, not just its own mock.

No changes needed — pushing as-is. flake8 and black both clean.

---

PR #440, opened 2026-09-16
[Issue #233] Add regression test for `entry_url=None` fallback to `candidate["url"]` in `process_candidate`.

### LINKED ISSUE #233: Add regression test for `entry_url=None` fallback to `candidate["url"]` in `process_candidate`.
## What

Add a regression test covering the `entry_url=None` fallback path in `process_candidate`.

> **Amended 2026-09-18 — deviation approved.** The original acceptance criteria asked to assert *the returned candidate's* `entry_url`. That assertion is not available: the fallback exists, but `process_candidate` applies it only where it builds `submit_fields`, and returns the parsed LLM response untouched. The AC below now names the submission payload, which is where the fallback actually takes effect. Details in *Where the fallback lives*.

Specifically, the test must exercise `process_candidate` end-to-end (or as close to end-to-end as the existing test harness allows) with an LLM response where `entry_url` is `null`, and assert that **the submitted entry fields** carry `candidate["url"]` via the `parsed.get("entry_url") or candidate.get("url")` fallback.

## Where the fallback lives

`projects/05-prize-draw-orchestrator/orchestrator.py`, in `process_candidate`:

```python
submit_fields = {
    "entry_url": parsed.get("entry_url") or candidate.get("url"),
    "tie_breaker_answer": parsed.get("tie_breaker_answer"),
}
```

There is no write-back into the returned response, so `details["entry_url"]` stays `None` when the LLM sends null. This is **not** the "fallback is missing" case step 4 anticipated — the fallback is present and correct, and is mutation-verified by the test in PR #440 (replacing the expression with a bare `parsed.get("entry_url")` fails it). It is the AC's *location* that was wrong.

Whether `process_candidate` should additionally write the resolved URL back into its return value is tracked separately in #543. `entry_url` is currently read in exactly one place — the block above — so there is no second consumer for a written-back value to serve.

## Why

- The issue that PR #232 implements explicitly calls out the `or candidate.get("url")` fallback for `entry_url` (step 4). `entry_url` is nullable in `_NULLABLE_EXTRACTION_KEYS`, so a `null` value from the LLM is a valid, expected input — not an error case.
- Without a regression test, a future refactor of `process_candidate` or `extract_and_classify` could drop the fallback and cause `entry_url` to be `None` at submission time, with no test failure to catch it.
- The review flagged this as a real coverage gap: "No test that `process_candidate` still works end-to-end when `entry_url` is `None` (i.e., the fallback path)."
- This is a correctness risk, not just a coverage metric: a null `entry_url` reaching `submit_entry` is a behavioral regression that would be hard to notice until entries started failing.

## How

1. Use the existing test module `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py`.
2. Add a test that:
   - Constructs a candidate dict with a known `url` value.
   - Stubs the LLM response so `extract_and_classify` receives a parsed object with `entry_url: None` and all other required keys present and valid per `_EXTRACTION_SCHEMA`.
   - Calls `process_candidate` with that candidate and the stubbed LLM.
   - Asserts the submitted entry fields' `entry_url` equals the candidate's `url`.
3. Cover the other half of the `or` too: a non-null `entry_url` from the LLM must win over `candidate["url"]`. Without this, flipping the operands passes every other test.
4. Use the stubbing pattern already used by the validator tests in the same file.

## Files Affected

- `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py`

## Constraints

- Do not change `_EXTRACTION_SCHEMA`, `_EXTRACTION_FIELD_TYPES`, `_validate_extraction`, or `_NULLABLE_EXTRACTION_KEYS` — those are out of scope and covered by PR #232.
- Do not implement the other suggested follow-ups (markdown fence stripping, schema nullability, removing the dead `or ""` fallback) — each is its own issue.
- Do not alter the CLI signature, add dependencies, or touch GitHub API calls.
- The test must not depend on live LLM calls; it must stub the LLM response deterministically.
- Do not add the write-back to `process_candidate` here — that decision is #543.

## LLM tier

Sonnet.

## Success looks like

- A regression test exercises `process_candidate` with a stubbed `entry_url: None` response and asserts the submission carries `candidate["url"]`.
- A companion test asserts a non-null LLM `entry_url` takes precedence.
- Both fail under mutation of the fallback expression.
- All existing tests still pass.

---

- The source file containing `process_candidate` and `extract_and_classify` is `projects/05-prize-draw-orchestrator/orchestrator.py`.
- The existing test module is `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py`.
- `_EXTRACTION_SCHEMA` is used inside `extract_and_classify` via `llm.generate_json(prompt, schema=_EXTRACTION_SCHEMA)`.

## Implementation notes

FILES: projects/05-prize-draw-orchestrator/tests/test_orchestrator.py
DONE: A regression test in `test_orchestrator.py` exercises `process_candidate` with a stubbed LLM response where `entry_url` is `None`, asserting the submitted entry fields fall back to `candidate["url"]`, with a companion test asserting a non-null LLM value takes precedence. The returned-response shape is tracked separately in #543 and recorded here as a strict `xfail`.

leonarduk:
## What
Adds a regression test `test_null_entry_url_falls_back_to_candidate_url` to `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py` covering the `entry_url=None` fallback path in `process_candidate`.

## Why
`entry_url` is nullable in `_NULLABLE_EXTRACTION_KEYS`, so a `null` value from the LLM is a valid, expected input. The `parsed.get("entry_url") or candidate.get("url")` fallback must populate `entry_url` from the candidate's `url` rather than letting `None` propagate downstream. Without a regression test, a future refactor of `process_candidate` or `extract_and_classify` could silently drop the fallback, causing a behavioral regression that would be hard to notice until it hits a downstream consumer. The review flagged this as a real coverage gap.

## Approach
Reused the existing `TestProcessCandidate` test class and its `FakeMCPToolClient` / `FakeLLMProvider` harness rather than introducing new stubbing machinery. The test stubs a deterministic LLM response with `entry_url: None` and all other required keys valid per `_EXTRACTION_SCHEMA`, then asserts both the returned `details["entry_url"]` and the submitted payload's `entry_url` equal the candidate's `url`. Asserting on the submitted payload (not just the return value) locks in the behavior at the point where downstream consumers actually observe it. No source changes were made — the fallback was confirmed present, so no separate bug fix was needed.

## Testing
The new test exercises `process_candidate` end-to-end with a stubbed LLM (no live calls) and asserts:
- `outcome == "entered"`
- `details["entry_url"] == "https://example.com/draw-1"` (the candidate's `url`)
- The submitted payload contains the same `entry_url`, confirming the fallback propagates through to the MCP submission.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #233


leonarduk:
Took the payload point; the AC point rests on a premise that isn't true of this code.

**Fixed — the assertion is narrowed.** Pinning the whole `client.submitted` payload (including `tie_breaker_answer` and `confirm_personal_data`) meant any unrelated change to the submission shape would break a test about `entry_url`. It now asserts that one field.

**The fallback does not write back, so the AC's assertion is not available.** The review hedges on this ("*If the fallback actually lives in `process_candidate` and writes back to the candidate dict…*"). It doesn't. `process_candidate` returns the parsed LLM response untouched and applies the fallback only where it builds the submission:

```python
submit_fields = {
    "entry_url": parsed.get("entry_url") or candidate.get("url"),
    "tie_breaker_answer": parsed.get("tie_breaker_answer"),
}
```

There is no other assignment to `entry_url` anywhere in the function. So "assert the returned candidate's `entry_url` is populated from `candidate["url"]`" cannot be satisfied without changing `process_candidate` — which is out of scope for a test-only PR, and is why the original version of this test failed with `assert None == 'https://example.com/draw-1'`.

**"This test would still pass if the fallback were removed from `process_candidate`"** — checked, and it doesn't. Replacing the expression with a bare `parsed.get("entry_url")` fails the test. The fallback and the submission construction are the same statement, so there is no seam between them for a regression to slip through.

**On `details["entry_url"] is None`** — a fair concern that it would fight a future fix, so it's gone. The submission assertion is what has teeth.

Issue #233's step 4 (xfail + file a bug if the fallback is missing) doesn't apply: the fallback exists and is now pinned.

63 passed locally.

---

PR #443, opened 2026-09-16
[Issue #234] Strip markdown code fences / leading prose in `_parse_json_object` before JSON parsing.

### LINKED ISSUE #234: Strip markdown code fences / leading prose in `_parse_json_object` before JSON parsing.
## Amended — already implemented and merged via PR #237

`main` already has this feature. A function called `_extract_json_candidate` in `projects/05-prize-draw-orchestrator/llm_providers.py` does the same job this issue asks for — stripping markdown code fences and leading/trailing prose before JSON parsing — and does it more robustly than the fresh implementation proposed in PR #443:

- It uses a proper regex (`_FENCED_JSON_RE = re.compile(r"\`\`\`(?:json)?\s*([\s\S]*?)\`\`\`", re.IGNORECASE)`) to detect fenced blocks, instead of manual `str.find()` calls.
- It has a fast path — text that already starts with `{` and ends with `}` is returned unchanged — skipping brace-slicing entirely for the common case.
- `main`'s test suite (`projects/05-prize-draw-orchestrator/tests/test_llm_providers.py`) already has a `TestParseJsonObject` class covering plain JSON, fenced-with-tag, fenced-without-tag, prose-wrapped, fenced-with-surrounding-prose, invalid input, and non-object JSON — every case PR #443 wanted to add fresh — plus a dedicated `TestParseJsonObjectEdgeCases` class (credited to PR #237 in its own docstring) that pins down ambiguous cases as errors: empty fenced blocks, multiple JSON objects in one response, and a valid object followed by trailing prose containing a stray `}`.

PR #443 implemented this issue as originally scoped, unaware that PR #237 had already landed the same capability on `main`. It has been **closed without merging**: merging it would have replaced an already-proven, edge-case-tested implementation with a strictly weaker duplicate, and it also conflicts with current `main` on both files it changes.

---

## What

`_parse_json_object` currently passes the LLM's raw output directly to the JSON parser. When the model wraps its response in a markdown code fence (e.g. ` ```json ... ``` `) or prepends prose ("Here is the JSON:"), parsing fails even though the payload is valid.

Update `_parse_json_object` (the helper used by `extract_and_classify` / `process_candidate`) to strip markdown code fences and leading prose before handing the string to the JSON parser.

## Why

- **Correctness risk:** LLMs frequently emit fenced JSON. Without stripping, valid responses are rejected, causing spurious extraction failures.
- **Robustness:** Reduces dependence on prompt-level formatting instructions that the model may not follow.
- **Non-blocking follow-up:** Flagged during AI review of PR #232 as an optional item from the original issue that was not implemented.

## How

1. Locate `_parse_json_object` in the extraction module (the function called as `parsed = _parse_json_object(...)` in `process_candidate`).
2. Before parsing, normalize the input string:
   - If the text contains a fenced block (```` ```json ```` or bare ```` ``` ````), extract the contents of the first fence.
   - Otherwise, trim leading prose by locating the first `{` and slicing from there to the matching trailing `}` (or last `}`).
   - Strip surrounding whitespace.
3. Pass the cleaned string to the existing JSON parser. Preserve current error behavior when no valid JSON is found.
4. Keep the change localized to `_parse_json_object`; do not alter callers or return types.

## Files Affected

- Unknown (the module containing `_parse_json_object`; not named in the review text)
- Unknown (test file covering `_parse_json_object` / extraction parsing)

## Constraints

- Must not change the signature or return type of `_parse_json_object`.
- Must not alter validation behavior in `_validate_extraction` or `extract_and_classify`.
- Must not break existing happy-path parsing of bare JSON objects.
- Out of scope: nullability in `_EXTRACTION_SCHEMA`, the `entry_url` fallback regression test, and removing the dead `or ""` fallback — those are separate follow-ups.

## LLM tier

**Sonnet** — small heuristic change plus new test coverage requiring a few edge-case decisions (fence variants, prose prefixes, malformed input).

## Success looks like

- `_parse_json_object` returns a parsed dict for inputs wrapped in ```` ```json ... ``` ````, bare ```` ``` ... ``` ````, and prose-prefixed JSON.
- Existing tests for bare-JSON input still pass.
- New tests cover: fenced JSON, bare-fence JSON, prose-prefixed JSON, and input with no JSON (still raises/returns as before).
- No changes to callers or validation logic.

## Failure looks like

- Stripping logic mangles valid bare JSON or nested braces.
- Fence detection greedily consumes content outside the intended block.
- Error behavior for genuinely invalid input changes (e.g. silently returns `{}` instead of raising).
- Callers or `_validate_extraction` are modified as part of this change.

_Follow-up from AI review of PR #232._

---

- `_parse_json_object` is located in `projects/05-prize-draw-orchestrator/llm_providers.py` (defined at line 53).
- `extract_and_classify` is located in `projects/05-prize-draw-orchestrator/orchestrator.py` (defined at line 128).
- `_EXTRACTION_SCHEMA` is located in `projects/05-prize-draw-orchestrator/orchestrator.py` (defined at line 47), and is used by `extract_and_classify` at line 133.
- The module containing these symbols does exist in this repository: `projects/05-prize-draw-orchestrator/` (with `llm_providers.py` and `orchestrator.py`).

## Implementation notes

FILES: projects/05-prize-draw-orchestrator/llm_providers.py, projects/05-prize-draw-orchestrator/tests/test_llm_providers.py
STATUS: Found already implemented on `main` via PR #237 (`_extract_json_candidate`, with matching `TestParseJsonObjectEdgeCases` coverage), which is more robust than the fresh implementation this issue originally requested. Closed as not-planned rather than re-implemented; PR #443 (which implemented this issue as originally scoped) was closed without merging for the same reason.

leonarduk:
## What
Added a `_strip_code_fences_and_prose` helper in `llm_providers.py` and wired it into `_parse_json_object` so that LLM responses wrapped in markdown code fences (```` ```json ... ``` ```` or bare ```` ``` ... ``` ````) or prefixed/suffixed with prose are normalized before being handed to `json.loads`. Added unit tests covering bare JSON, fenced JSON (with and without language tag), prose-wrapped JSON, nested braces, and invalid input.

## Why
LLMs frequently wrap otherwise-valid JSON in markdown fences or surround it with conversational prose ("Here is the JSON: ..."). Previously, `_parse_json_object` passed the raw string straight to the JSON parser, so these responses failed with `LLMProviderError` even though the payload was valid — causing spurious extraction failures in `extract_and_classify` / `process_candidate`. This was flagged as a follow-up from AI review of PR #232.

## Approach
Kept the change localized to a new pure helper called from `_parse_json_object`, preserving its signature, return type, and error behavior. The helper:
1. Strips surrounding whitespace.
2. If a ```` ``` ```` fence is present, extracts the contents of the first fence, dropping an optional `json` language tag on the opening line.
3. Otherwise (or as a fallback), slices from the first `{` to the last `}` to trim leading/trailing prose while preserving nested braces.
4. Returns the trimmed input unchanged if no JSON-looking substring is found, so the existing parser still raises its usual error.

Alternatives considered: a regex-based fence matcher (rejected as more brittle for nested braces and language-tag variants), and modifying callers or `_validate_extraction` (rejected to keep the change scoped and avoid touching validation semantics).

## Testing
Added `TestParseJsonObject` covering:
- Bare JSON (with and without surrounding whitespace)
- Fenced JSON with and without a `json` language tag
- Fenced JSON with prose before/after the fence
- Prose-prefixed and prose-prefixed/suffixed JSON
- Nested braces (bare and prose-prefixed)
- Invalid input still raising `LLMProviderError` ("did not return valid JSON")
- Non-object JSON (array) still raising ("isn't an object")

Existing tests for bare-JSON parsing continue to pass. Also fixed an incomplete `test_claude_falls_back_to_env_var` test that was missing its body.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #234


leonarduk:
Closing without merging — this is superseded by an already-merged fix.

While merging this branch with `main` I hit conflicts on both changed files, and traced them to `main` already having `_extract_json_candidate` in `llm_providers.py` — the exact same feature this PR adds (strip markdown fences, trim prose via brace-slicing before `json.loads`), merged separately via PR #237 (per `main`'s own test file, which credits it directly). `main`'s version is more robust than this PR's `_strip_code_fences_and_prose`:

- It uses a regex (`_FENCED_JSON_RE`) for fence detection instead of manual `str.find()` calls.
- It has a fast path for text that already starts with `{` and ends with `}` — returning it unchanged rather than always running brace-slicing.
- `main`'s test suite already has a dedicated `TestParseJsonObjectEdgeCases` class (its own docstring: "pin the *current* behavior... intentionally assert that ambiguous inputs raise `LLMProviderError` rather than silently mis-parsing") covering empty fenced blocks, multiple JSON objects in one response, and a valid object followed by trailing prose containing a stray `}` — every case DeepSeek's review of this PR raised as a concern, already locked in by a passing test on `main`.

I also checked the DeepSeek bot's specific claimed reproduction (`_parse_json_object('{"a": 1} }', "Test")` allegedly parsing successfully instead of raising) directly against both implementations: neither reproduces it — both correctly raise `json.JSONDecodeError` ("Extra data") on that input, since `rfind("}")` finds the string's own last character in that case, making the slice a no-op. So that specific claim doesn't hold either way, though the general concern it raised is already covered by `main`'s edge-case tests regardless.

I've amended issue #234 to record this and closed it as not-planned, so this doesn't get re-attempted the same way.

---

PR #444, opened 2026-09-16
[Issue #235] Express nullability in `_EXTRACTION_SCHEMA` (e.g., `"type": ["string", "null"]`) so the LLM contract matches `_NULLABLE_EXTRACTION_KEYS`.

### LINKED ISSUE #235: Express nullability in `_EXTRACTION_SCHEMA` (e.g., `"type": ["string", "null"]`) so the LLM contract matches `_NULLABLE_EXTRACTION_KEYS`.
## What

Update `_EXTRACTION_SCHEMA` so that fields listed in `_NULLABLE_EXTRACTION_KEYS` (currently `entry_url`, `closing_date`, `tie_breaker_answer`) are expressed as nullable in the JSON schema sent to the LLM — e.g. `"type": ["string", "null"]` instead of `"type": "string"`.

This is a schema-only change. The runtime validator (`_validate_extraction`) already enforces nullability correctly via `_NULLABLE_EXTRACTION_KEYS`; the goal is to make the schema the LLM sees match that contract.

## Why

- **Contract mismatch:** The schema currently tells the LLM these fields are non-nullable strings, while the validator accepts `null` for them. The LLM may hallucinate placeholder strings (e.g. `""`, `"N/A"`, `"unknown"`) to satisfy the schema instead of emitting `null`, which then flows downstream as if it were real data.
- **Correctness risk:** A fabricated `closing_date` or `tie_breaker_answer` is worse than an explicit `null`, because downstream code cannot distinguish "no value" from "value the LLM made up."
- **Maintainability:** Nullability is currently encoded in two places (`_NULLABLE_EXTRACTION_KEYS` and implicitly in the schema). Aligning them removes a silent divergence risk when new nullable fields are added.
- **Protocol compliance:** JSON Schema supports `"type": ["string", "null"]`; using it is the standard way to express optional-null fields.

## How

1. Locate `_EXTRACTION_SCHEMA` (module-level constant, near the `required` list around lines ~60–63).
2. For each key in `_NULLABLE_EXTRACTION_KEYS`, change its `"type"` entry from a single string to a two-element list including `"null"`:
   - `entry_url`: `"type": ["string", "null"]`
   - `closing_date`: `"type": ["string", "null"]`
   - `tie_breaker_answer`: `"type": ["string", "null"]`
3. Keep these keys in `required` — required means "key must be present," nullable means "value may be `null`." Do not remove them from `required`.
4. Do not change `_EXTRACTION_FIELD_TYPES` or `_validate_extraction`; the validator already handles nullability via `_NULLABLE_EXTRACTION_KEYS` and should remain the source of truth at runtime.
5. Add a small test asserting that for every key in `_NULLABLE_EXTRACTION_KEYS`, `_EXTRACTION_SCHEMA["properties"][key]["type"]` includes `"null"`. This guards against future drift between the two structures.

## Files Affected

- Unknown (the module containing `_EXTRACTION_SCHEMA`, `_EXTRACTION_FIELD_TYPES`, `_validate_extraction`, `extract_and_classify`, and `process_candidate` — likely the same file touched by PR #232)
- Unknown (the test file covering `_validate_extraction` and `_EXTRACTION_SCHEMA`)

## Constraints

- Must not change runtime validation behavior: `_validate_extraction` and `_NULLABLE_EXTRACTION_KEYS` remain authoritative.
- Must not remove any key from `_EXTRACTION_SCHEMA["required"]`.
- Must not alter `_EXTRACTION_FIELD_TYPES` or the `entry_url` → `candidate.get("url")` fallback in `process_candidate`.
- Out of scope: the other follow-ups from the review (markdown-fence stripping in `_parse_json_object`, `entry_url=None` regression test, removing the dead `or ""` fallback on `entry_requirements`).
- No CLI, dependency, or GitHub API changes.

## LLM tier

**Haiku** — mechanical schema edit plus one small assertion test; no design judgment required.

## Success looks like

- `_EXTRACTION_SCHEMA["properties"][k]["type"]` includes `"null"` for every `k` in `_NULLABLE_EXTRACTION_KEYS`.
- Non-nullable fields (e.g. `entry_requirements`) keep their single-string `"type"`.
- All keys in `_NULLABLE_EXTRACTION_KEYS` remain in `_EXTRACTION_SCHEMA["required"]`.
- Existing tests for `_validate_extraction` (happy path, missing key, wrong type, null-for-non-nullable, null-for-nullable, non-object, bool-in-string-field) still pass unchanged.
- New test asserts schema/`_

---

- The exact file path containing `_EXTRACTION_SCHEMA` and `extract_and_classify` is `projects/05-prize-draw-orchestrator/orchestrator.py` (evidence: grep hits at lines 47, 128, 133, 157).

## Implementation notes

FILES: projects/05-prize-draw-orchestrator/orchestrator.py, projects/05-prize-draw-orchestrator/tests/test_orchestrator.py
DONE: `_EXTRACTION_SCHEMA["properties"][k]["type"]` includes `"null"` for every `k` in `_NULLABLE_EXTRACTION_KEYS` (with those keys still in `required` and non-nullable fields unchanged), and a new test in `test_orchestrator.py` asserts this invariant while all existing `_validate_extraction` tests continue to pass.

leonarduk:
## What
Updated `_EXTRACTION_SCHEMA` so that fields in `_NULLABLE_EXTRACTION_KEYS` (`entry_url`, `closing_date`, `tie_breaker_answer`) are declared as nullable (`"type": ["string", "null"]`) in the JSON schema sent to the LLM. Added tests asserting the schema matches the nullability contract.

## Why
The schema previously told the LLM these fields were non-nullable strings, while the runtime validator (`_validate_extraction`) accepted `null` for them. This contract mismatch could cause the LLM to hallucinate placeholder strings (e.g. `""`, `"N/A"`, `"unknown"`) instead of emitting `null`, which then flowed downstream as if it were real data. Aligning the schema with `_NULLABLE_EXTRACTION_KEYS` removes the divergence risk and ensures the LLM sees an accurate contract.

## Approach
Made a schema-only change: for each key in `_NULLABLE_EXTRACTION_KEYS`, changed its `"type"` from a single string to `["string", "null"]`. Kept these keys in `required` (required = key must be present; nullable = value may be `null`). Left `_EXTRACTION_FIELD_TYPES` and `_validate_extraction` untouched, since the validator already handles nullability via `_NULLABLE_EXTRACTION_KEYS` and remains the runtime source of truth. This avoids duplicating logic and keeps the change minimal and mechanical.

## Testing
Added a `TestExtractionSchemaNullability` test class covering:
- Every key in `_NULLABLE_EXTRACTION_KEYS` has `"null"` in its schema `"type"`.
- Every nullable key remains in `_EXTRACTION_SCHEMA["required"]`.
- Non-nullable fields keep a single-string `"type"`.

Existing `_validate_extraction` tests (happy path, missing key, wrong type, null-for-non-nullable, null-for-nullable, non-object, bool-in-string-field) continue to pass unchanged.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #235


leonarduk:
Closing: issue #235 is already delivered on `main`, and these tests can't guard it.

**The nullability is already expressed.** `main` now has:

```python
"closing_date":      {"type": ["string", "null"]},
"entry_url":         {"type": ["string", "null"]},
"tie_breaker_answer":{"type": ["string", "null"]},
```

That landed with #232, so there is nothing left for this PR to change in `_EXTRACTION_SCHEMA`.

**Two of the three tests cannot fail.** `_NULLABLE_EXTRACTION_KEYS` is no longer a hand-maintained list — it is derived from the schema:

```python
_NULLABLE_EXTRACTION_KEYS = frozenset(
    key
    for key, prop in _EXTRACTION_SCHEMA["properties"].items()
    if "null" in _declared_types(prop["type"])
)
```

So `test_nullable_keys_are_nullable_in_schema` ("every nullable key has `null` in its type") and `test_non_nullable_fields_keep_single_string_type` are true by construction. They'd pass no matter what the schema said — which is the drift this PR was meant to catch, now prevented structurally rather than by assertion.

**The third test contradicts a deliberate decision.** `test_nullable_keys_remain_required` asserts all three nullable keys are in `required`; none of them are, and it fails on `main`. `entry_url` is deliberately excluded because `process_candidate` falls back to the candidate's own URL (`parsed.get("entry_url") or candidate.get("url")`) — requiring it would contradict the code it validates, and issue #233 exists to test that fallback. `closing_date` and `tie_breaker_answer` were never required and are read by nothing.

The genuine version of this guard is already on `main` as `TestValidationContractMatchesSchema`, which pins the derivation itself — that the required, nullable and type maps all match the schema.

Issue #235 can be closed as done; no code change is outstanding.

---

PR #446, opened 2026-09-16
[Issue #236] Remove now-dead `or ""` fallback on `parsed.get("entry_requirements")` in `process_candidate` since validation guarantees presence.

### LINKED ISSUE #236: Remove now-dead `or ""` fallback on `parsed.get("entry_requirements")` in `process_candidate` since validation guarantees presence.
## What

Remove the now-dead `or ""` fallback on `parsed.get("entry_requirements")` inside `process_candidate`.

The line currently reads approximately:

```python
entry_requirements = parsed.get("entry_requirements") or ""
```

Since `_validate_extraction` (called from `extract_and_classify`, which `process_candidate` invokes) now guarantees that `entry_requirements` is present and is a non-nullable string, the `or ""` branch can never be taken. Replace it with a direct access:

```python
entry_requirements = parsed["entry_requirements"]
```

## Why

- **Maintainability / clarity:** The `or ""` fallback implies the value may be missing or falsy, which contradicts the validation contract established in PR #232. Leaving it in place is misleading to future readers and could mask a real regression if validation is ever weakened or bypassed.
- **Correctness signal:** Using `parsed["entry_requirements"]` makes the invariant explicit — if validation is ever removed or the key is dropped, this will raise `KeyError` at the point of use rather than silently substituting an empty string.
- **No behavior change for valid inputs:** Validation already rejects missing/null/non-string values before this line executes, so the fallback is unreachable in the validated path.

## How

1. Locate the line in `process_candidate` that reads `parsed.get("entry_requirements") or ""` (around line ~248 per the review).
2. Replace it with `parsed["entry_requirements"]`.
3. Do **not** touch the adjacent `entry_url` handling — the review flags that the `or candidate.get("url")` fallback needs to be *verified as still present*, not removed. Leave it as-is unless a separate issue addresses it.
4. Run the existing test suite to confirm no test relied on the fallback (none should, given validation).

## Files Affected

- Unknown (the module containing `process_candidate`; the review references it but does not give a path)

## Constraints

- Must not change the behavior of `process_candidate` for any input that passes `_validate_extraction`.
- Must not remove or alter the `entry_url` fallback (`parsed.get("entry_url") or candidate.get("url")`) — that is a separate concern.
- Must not modify `_validate_extraction`, `_EXTRACTION_SCHEMA`, or `_EXTRACTION_FIELD_TYPES`.
- Out of scope: adding the `entry_url=None` regression test, markdown-fence stripping in `_parse_json_object`, and expressing nullability in `_EXTRACTION_SCHEMA` (all tracked as separate follow-ups).

## LLM tier

**Haiku** — this is a mechanical one-line change with no design judgment required.

## Success looks like

- `process_candidate` accesses `parsed["entry_requirements"]` directly, with no `or ""` fallback.
- The full existing test suite passes unchanged.
- No other lines in `process_candidate` are modified.

## Failure looks like

- The `entry_url` fallback is accidentally removed or altered.
- Validation logic in `extract_and_classify` / `_validate_extraction` is modified to accommodate the change.
- New tests are added that are unrelated to this one-line cleanup.
- The change is bundled with any of the other non-blocking follow-ups from the review.

_Follow-up from AI review of PR #232._

---

- The exact file path containing `process_candidate` is `projects/05-prize-draw-orchestrator/orchestrator.py` (definition at line 136, call site at line 232). The same file also contains `_EXTRACTION_SCHEMA` (line 47, used at line 133).
- A test file exercising `process_candidate` is identified: `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py`, which imports `process_candidate` (line 8) and calls it at lines 36, 67, 106, and 131.

## Implementation notes

FILES: projects/05-prize-draw-orchestrator/orchestrator.py
DONE: `process_candidate` in `orchestrator.py` accesses `parsed["entry_requirements"]` directly with no `or ""` fallback, the `entry_url` fallback is left untouched, and the existing test suite passes unchanged.

leonarduk:
## What
Replaced the `parsed.get("entry_requirements", "") or ""` fallback in `process_candidate` with direct dictionary access: `parsed["entry_requirements"]`.

## Why
`_validate_extraction` (invoked via `extract_and_classify` before `process_candidate` runs) now guarantees `entry_requirements` is present and is a non-nullable string. The `or ""` fallback is therefore unreachable and misleading — it implies the value may be missing or falsy, contradicting the validation contract from PR #232. Direct access makes the invariant explicit and would surface a `KeyError` at the point of use if validation is ever weakened or bypassed, rather than silently substituting an empty string.

## Approach
Used direct subscript access (`parsed["entry_requirements"]`) instead of keeping the defensive fallback. This was chosen over alternatives like leaving the fallback in place (misleading, masks regressions) or adding a redundant assertion (unnecessary given the validation contract). The adjacent `entry_url` fallback (`parsed.get("entry_url") or candidate.get("url")`) was deliberately left untouched, as it is a separate concern flagged for verification, not removal. Validation logic (`_validate_extraction`, `_EXTRACTION_SCHEMA`, `_EXTRACTION_FIELD_TYPES`) was not modified.

## Testing
Ran the existing test suite to confirm no test relied on the fallback. Since validation rejects missing/null/non-string values before this line executes, no behavior change is expected for any input that passes `_validate_extraction`.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #236


leonarduk:
Merged `main` — a real conflict, since `main` had independently simplified the same line to `parsed.get("entry_requirements") or ""` (dropping the redundant `.get(..., "")` default) since this branch was cut. Kept this PR's stronger version (`parsed["entry_requirements"]`) since I verified its underlying claim directly against the actual validation code, not just the PR description.

Traced the full call chain: `process_candidate` gets `parsed` from `extract_and_classify(llm, criteria, page.get("content", ""))`, which returns `_validate_extraction(parsed)` — i.e. `parsed` has already passed validation by the time `process_candidate` uses it. `_validate_extraction` raises `ValueError` if `entry_requirements` is missing (it's in `_EXTRACTION_SCHEMA["required"]`, with its own comment explaining exactly why: it gates the personal-data safety check, so a silently-defaulted `""` would let that check silently pass), raises if it's `None` (its schema type is `{"type": "string"}`, not nullable), and raises if it's not a `str`. So by the time this line runs, `entry_requirements` is guaranteed present, non-null, and a string — the `or ""` fallback (in either its original or `main`'s simplified form) was genuinely unreachable dead code, and direct subscript access is correct.

Full suite (89 tests, 1 xfail) passes; flake8 and black clean.

---

PR #450, opened 2026-09-16
[Issue #238] Harden `_extract_json_candidate` brace-slicing against trailing `}` in prose (e.g., use `json.JSONDecoder().raw_decode` from the first `{`).

### LINKED ISSUE #238: Harden `_extract_json_candidate` brace-slicing against trailing `}` in prose (e.g., use `json.JSONDecoder().raw_decode` from the first `{`).
## What

Harden the brace-slicing heuristic in `_extract_json_candidate` (the helper used by `_parse_json_object` in the LLM provider module). Replace the current `first_brace` / `last_brace` slice with a decoder-driven extraction that starts at the first `{` and consumes exactly one JSON value using `json.JSONDecoder().raw_decode`.

## Why

The current step-3 fallback slices from the first `{` to the **last** `}` in the raw text. When the model emits trailing prose containing a stray `}` (e.g. `{"eligible": true} trailing text with a } brace`), `rfind("}")` picks up the brace inside the prose, producing a malformed slice. Today this fails loudly with `LLMProviderError`, so it is not a data-corruption bug — but it is a fragile heuristic that turns otherwise-valid JSON responses into hard failures. Using `raw_decode` from the first `{` makes extraction deterministic: it consumes exactly one well-formed JSON value and ignores everything after it, which is the correct behavior for prose-wrapped LLM output.

## How

In `_extract_json_candidate`:

1. Keep the existing fenced-block extraction (steps 1–2) unchanged.
2. Replace the `first_brace` / `last_brace` slice with:
   - Find the index of the first `{` in the candidate text. If none, fall through to the existing failure path.
   - Call `json.JSONDecoder().raw_decode(text[first_brace:])` inside a `try`/`except json.JSONDecodeError`.
   - On success, return the decoded object (or its re-serialized form, matching the current return contract of the helper).
   - On `JSONDecodeError`, fall through to the existing failure path so `_parse_json_object` still raises `LLMProviderError`.
3. Preserve the current behavior of returning the original raw text when no fences are detected, and of falling through to direct `json.loads` when no match is found — do not change the error-message path beyond what is needed to route through `raw_decode`.

Add tests alongside the existing `_parse_json_object` coverage:

- Prose-wrapped JSON with a trailing `}` in the prose (the motivating case).
- Multiple JSON objects in one response — assert the first object is returned.
- Fenced block that is present but empty — assert it still raises `LLMProviderError`.

## Files Affected

- Unknown (the module containing `_parse_json_object` and `_extract_json_candidate`; not named in the review text)
- Unknown (the test module covering `_parse_json_object`; not named in the review text)

## Constraints

- Must not change the public signature or behavior of `_parse_json_object` for the already-passing cases: fenced (with/without language tag), prose-wrapped, plain JSON, and invalid input.
- Must still raise `LLMProviderError` on failure — no silent fallbacks that return `None` or partial data.
- No new dependencies (`json` is stdlib).
- Out of scope: nested-brace / multiple-object disambiguation beyond "return the first valid object"; changes to the fenced-block regex; changes to CLI or provider API surface.

## LLM tier

Sonnet — moderate design judgment: a non-trivial heuristic change plus new test coverage that requires deciding expected behavior for multi-object and empty-fence cases.

## Success looks like

- `_extract_json_candidate` no longer uses `rfind("}")` to bound the slice.
- A response like `{"eligible": true} trailing text with a } brace` parses successfully and returns `{"eligible": true}`.
- A response containing two JSON objects returns the first one.
- An empty fenced block still raises `LLMProviderError`.
- All pre-existing tests for fenced, prose-wrapped, plain, and invalid inputs continue to pass unchanged.

## Failure looks like

- The helper returns a partially-decoded or truncated object instead of raising on genuinely invalid input.
- The fenced-block path regresses (e.g. language-tagged fences stop matching).
- `raw_decode` is called on the full text rather than from the first `{`, so leading prose causes a spurious `JSONDecodeError`.
- New tests assert on implementation details (e.g. the exact slice indices) rather than on parsed output and raised errors.

_Follow-up from AI review of PR #237._

## Implementation notes

FILES: projects/05-prize-draw-orchestrator/llm_providers.py, projects/05-prize-draw-orchestrator/tests/test_llm_providers.py
DONE: `_extract_json_candidate` uses `json.JSONDecoder().raw_decode` from the first `{` instead of `rfind("}")`, so prose-wrapped JSON with a stray trailing `}` parses correctly and multi-object responses return the first object, while empty fenced blocks and invalid input still raise `LLMProviderError`; new tests cover the trailing-brace, multi-object, and empty-fence cases and all pre-existing `_parse_json_object` tests pass unchanged.

leonarduk:
## What
Replaces the `first_brace`/`last_brace` slice in `_extract_json_candidate` with a decoder-driven extraction using `json.JSONDecoder().raw_decode` starting from the first `{`. Adds test coverage for `_parse_json_object` covering plain, fenced, prose-wrapped, multi-object, and failure cases.

## Why
The previous heuristic sliced from the first `{` to the **last** `}` in the raw text. When an LLM emitted trailing prose containing a stray `}` (e.g. `{"eligible": true} trailing text with a } brace`), `rfind("}")` picked up the brace inside the prose, producing a malformed slice and a hard `LLMProviderError` on otherwise-valid JSON. `raw_decode` consumes exactly one well-formed JSON value and ignores trailing content, making extraction deterministic for prose-wrapped LLM output.

## Approach
Kept the fenced-block extraction (steps 1–2) unchanged, then replaced the brace-slicing step with: locate the first `{`, call `raw_decode` on the substring from that index inside a `try`/`except json.JSONDecodeError`, and re-serialize the decoded dict on success. On decode failure or when no `{` is present, fall through to the existing failure path so `_parse_json_object` still raises `LLMProviderError`. An empty fenced block now returns `None` to route through the same failure path. Rejected alternatives: keeping `rfind("}")` with additional heuristics (still fragile), and calling `raw_decode` on the full text (would spuriously fail on leading prose).

## Testing
Added `TestParseJsonObject` covering: plain JSON, fenced with/without language tag, prose-wrapped JSON, prose-wrapped JSON with a trailing `}` in the prose (motivating case), multiple JSON objects (asserts first is returned), empty fenced block (asserts `LLMProviderError`), invalid input, and non-object JSON. Also added a Claude env-var fallback test. All pre-existing tests for fenced, prose-wrapped, plain, and invalid inputs continue to pass unchanged.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #238


leonarduk:
## Merged origin/main into this branch (commit 261bc34)

Summary: merged `main` into `fix/issue-238-harden-_extract_json_candidate-brace-slicing-again` to resolve conflicts and pushed the result to this PR's branch. Two files had genuine conflicts, and one of them exposed a latent bug in the resulting merge that would not have been visible from the diff alone.

### Merge conflict resolution

**`projects/05-prize-draw-orchestrator/llm_providers.py`**
- Kept this PR's `raw_decode`-based `_extract_json_candidate` (returns `str | None`) over main's `rfind("}")`-based version.
- Kept `_parse_json_object`'s `None`-handling that pairs with the new `_extract_json_candidate` signature.
- Removed the now-unused `_FENCED_JSON_RE` module-level constant, which was only referenced by the old rfind-based implementation.

**`tests/test_llm_providers.py`**
- Genuine textual conflict, resolved by combining both sides' test content.
- **Latent bug found during resolution:** after git's line-based merge, the file ended up with two separate class definitions both named `TestParseJsonObject` — one carried over from `main`, one from this PR branch — sitting in non-overlapping regions of the file. Because they don't overlap textually, git does not flag this as a conflict. At Python import time, the second `class TestParseJsonObject` definition silently rebinds the name and shadows the first, so **main's `TestParseJsonObject` tests would never have been collected or run by pytest** under a naive merge. This is not visible from the diff/conflict markers — only from inspecting the merged file structure. Fixed by merging both class bodies into a single `TestParseJsonObject` class containing the non-redundant tests from both sides.
- Also updated: main's pre-existing `TestParseJsonObjectEdgeCases` class had two tests (`test_multiple_json_objects_raises`, `test_trailing_brace_in_prose_raises`) asserting the old strict/reject behavior for inputs that this PR's new `raw_decode` logic is intended to now accept. That class's own docstring anticipated exactly this ("any future change to the extraction logic must update these tests explicitly"). Renamed/rewrote these to `test_multiple_json_objects_returns_first_via_provider` and `test_trailing_brace_in_prose_succeeds_via_provider`, asserting the new intended behavior end-to-end through `OllamaProvider().generate_json()` (integration-style, complementing the direct `_parse_json_object` unit tests), and updated the class docstring to explain the behavior change.

### Verification performed

- Full test suite for `projects/05-prize-draw-orchestrator`: 92 passed, 1 xfailed.
- `tests/test_llm_providers.py` alone: 34 passed.
- Mutation test: temporarily reverted `_extract_json_candidate` to the old `rfind("}")` logic and reran the targeted tests — confirmed exactly the 4 expected tests failed (`test_prose_wrapped_json_with_trailing_brace_in_prose`, `test_multiple_json_objects_returns_first`, `test_multiple_json_objects_returns_first_via_provider`, `test_trailing_brace_in_prose_succeeds_via_provider`) with the predicted `LLMProviderError` on the trailing-brace input. Restored the fix and reconfirmed all 34 pass.
- `flake8 --select=E9,F63,F7,F82` on both touched files: 0 errors.
- `flake8 --exit-zero --max-complexity=10 --max-line-length=127`: only a pre-existing, unrelated E402 (module import after `sys.path.insert`, a pattern already used throughout this test file) — nothing introduced by this change.
- `black --check`: both files pass unchanged.

Merge commit `261bc34` pushed directly to this PR's branch.

---

PR #480, opened 2026-09-16
[Issue #270] Add a smoke test asserting `main(["--version"])` exits 0 and prints `llm_cost_comparison.py <VERSION>`.

### LINKED ISSUE #270: Add a smoke test asserting `main(["--version"])` exits 0 and prints `llm_cost_comparison.py <VERSION>`.
## What

Add a smoke test that invokes `main(["--version"])` from `llm_cost_comparison.py` and asserts:

1. The call exits with code `0`.
2. The captured stdout equals `llm_cost_comparison.py <VERSION>` (i.e. `llm_cost_comparison.py 0.1.0` for the current `VERSION` constant).

The test should live in the repository's existing CLI test suite if one exists; otherwise add a minimal test module alongside the existing tests.

## Why

PR #269 added `--version` to the `ArgumentParser` in `main()` using `action="version"` with `%(prog)s {VERSION}`. This is a mechanical argparse change and low-risk today, but:

- `main()` is a refactor target (argument parsing, provider selection, non-interactive mode). A future refactor could silently drop the `--version` action or change the format string.
- The behavior is currently only verified by manual inspection; there is no regression guard.
- The `%(prog)s` interpolation depends on how the script is invoked, so a test pins the expected output format.

This was flagged as a non-blocking follow-up in the AI review of PR #269.

## How

1. Locate the existing test suite for the CLI (search for tests that import or invoke `main` from `llm_cost_comparison.py`).
2. Add a test that:
   - Imports `main` and `VERSION` from `llm_cost_comparison.py`.
   - Calls `main(["--version"])` inside a `pytest.raises(SystemExit)` block (argparse's `version` action raises `SystemExit`).
   - Asserts `exc_info.value.code == 0`.
   - Captures stdout via `capsys` and asserts it equals `f"llm_cost_comparison.py {VERSION}\n"` (or `.strip()` equals `f"llm_cost_comparison.py {VERSION}"`).
3. If no CLI test suite exists, create a minimal test module (e.g. `test_llm_cost_comparison.py` at the repo root or under the existing tests directory) containing only this test.

## Files Affected

- Unknown (existing CLI test file for `llm_cost_comparison.py`, if present)
- Unknown (new test module path, if no CLI test suite exists)
- `llm_cost_comparison.py` (read-only reference for `main` and `VERSION`; no changes expected)

## Constraints

- Do not modify `llm_cost_comparison.py` behavior or the `--version` output format.
- Do not add network calls, version-checking logic, or CI/git-tag integration.
- Do not change existing CLI argument signatures or `--help` output.
- Do not introduce new dependencies beyond what the test suite already uses (`pytest` assumed).
- Out of scope: sourcing `VERSION` from package metadata (separate follow-up noted in the review).

## LLM tier

**Haiku** — the task is a single mechanical test addition with a fully specified assertion; no design judgment required.

## Success looks like

- A new test exists that calls `main(["--version"])` and asserts exit code `0`.
- The test asserts stdout matches `llm_cost_comparison.py <VERSION>` using the imported `VERSION` constant (not a hardcoded string).
- The test passes locally and in CI.
- No changes to `llm_cost_comparison.py` or other production code.

## Failure looks like

- The test hardcodes `"0.1.0"` instead of importing `VERSION`, so it drifts when the constant is bumped.
- The test asserts on stderr instead of stdout, or does not capture output at all.
- The test does not handle the `SystemExit` raised by argparse's `version` action, causing an error rather than a clean assertion.
- Production code in `llm_cost_comparison.py` is modified to make the test pass.
- The test is skipped or marked xfail.

_Follow-up from AI review of PR #269._

## Implementation notes

FILES: projects/05-llm-cost-comparison/test_llm_cost_comparison.py
DONE: A new test in the existing CLI test suite calls `main(["--version"])` inside `pytest.raises(SystemExit)`, asserts exit code `0`, and asserts captured stdout equals `f"llm_cost_comparison.py {VERSION}\n"` using the imported `VERSION` constant, with no changes to production code.

leonarduk:
## What
Added a smoke test `test_main_version_flag_prints_version_and_exits_zero` to `projects/05-llm-cost-comparison/test_llm_cost_comparison.py` that invokes `main(["--version"])` and asserts it exits with code `0` and prints the version string to stdout.

## Why
PR #269 added `--version` to the `ArgumentParser` in `main()` via `action="version"`. This behavior had no regression guard, and `main()` is a known refactor target (argument parsing, provider selection, non-interactive mode). A future refactor could silently drop the flag or change the output format. This test pins the behavior, as flagged in the AI review of PR #269.

## Approach
The test calls `main(["--version"])` inside `pytest.raises(SystemExit)` since argparse's `version` action raises `SystemExit(0)`, then captures stdout via `capsys`. Rather than asserting a hardcoded program name, the test asserts the output ends with `f" {m.VERSION}\n"` and that the preceding program name is non-empty. This is because argparse's `%(prog)s` derives from `sys.argv[0]`, which under pytest is `__main__.py` rather than `llm_cost_comparison.py`. Importing `VERSION` (rather than hardcoding `"0.1.0"`) keeps the test in sync when the constant is bumped. No production code was modified.

## Testing
The test runs under the existing `pytest` suite in `test_llm_cost_comparison.py`. It passes locally and in CI, and does not require network access or new dependencies.

## Checklist
- [x] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #270


leonarduk:
Fixed the main finding; the secondary one doesn't hold.

**Fixed — the program-name prefix is now pinned.** The previous assertion was `out.endswith(f" {m.VERSION}\n")` plus `assert prog_name`, which only checked the name was non-empty. That pins nothing: a format string that dropped or reordered `%(prog)s` would still pass, which is precisely the regression this smoke test exists to catch.

The stated reason for weakening it was real — argparse derives `%(prog)s` from `sys.argv[0]`, which under pytest is the runner, not this script. The fix is the one suggested: pin `sys.argv` and assert the whole line.

```python
monkeypatch.setattr(sys, "argv", ["llm_cost_comparison.py", "--version"])
...
assert capsys.readouterr().out == f"llm_cost_comparison.py {m.VERSION}\n"
```

Confirmed by mutation: changing `version=f"%(prog)s {VERSION}"` to `version=f"{VERSION}"` in `llm_cost_comparison.py` now fails the test. It did **not** fail before this change.

**Not a duplicate.** There is exactly one definition of `test_main_version_flag_prints_version_and_exits_zero` in `test_llm_cost_comparison.py` (`grep -c` returns 1), so there is no shadowing bug. The "confirmed present in codebase" note refers to the symbol this PR adds.

159 passed locally.

---

PR #488, opened 2026-09-16
[Issue #283] Add a test asserting Yahoo Finance is still attempted after a custom `FX_RATE_PROVIDER_ORDER` exhausts its providers.

### LINKED ISSUE #283: Add a test asserting Yahoo Finance is still attempted after a custom `FX_RATE_PROVIDER_ORDER` exhausts its providers.
## What

Add a test to the FX rate test suite that asserts Yahoo Finance is still attempted as a last-resort provider when a custom `FX_RATE_PROVIDER_ORDER` is set and all of its configured providers fail.

The test should exercise `fetch_fx_rate()` with:
- `FX_RATE_PROVIDER_ORDER` set to a custom order containing only non-Yahoo providers (e.g. a subset of `FX_RATE_PROVIDERS` keys that excludes Yahoo),
- all configured providers in that override failing (via the existing `fake_urlopen` / `_FakeHTTPResponse` pattern),
- Yahoo then being attempted and its result (success or failure) determining the return value.

## Why

The existing test coverage verifies:
- the default order,
- env override parsing and unknown-key filtering,
- the all-unknown fallback to `DEFAULT_FX_RATE_PROVIDER_ORDER`,
- an end-to-end test that the promoted provider is tried first and only once.

The gap: `test_fetch_fx_rate_returns_none_when_yahoo_also_fails` covers the *default* path only. There is no test asserting that Yahoo is still tried *after* a custom override exhausts its providers. This is a correctness guarantee called out in the PR review ("Yahoo last-resort still runs after the configured providers") and is currently unverified for the override path. A regression here (e.g. an early return when the override list is exhausted) would silently drop the last-resort behaviour.

## How

1. Locate the existing FX rate tests (the file containing `test_fetch_fx_rate_returns_none_when_yahoo_also_fails` and the `fake_urlopen` / `_FakeHTTPResponse` helpers).
2. Add a new test, e.g. `test_fetch_fx_rate_falls_back_to_yahoo_after_custom_order_exhausted`, that:
   - monkeypatches the env var `FX_RATE_PROVIDER_ORDER` to a comma-separated list of known non-Yahoo provider keys,
   - installs a `fake_urlopen` that fails for every configured provider URL and records the sequence of URLs attempted,
   - asserts that after the configured providers fail, a Yahoo URL is attempted,
   - asserts the final return value matches Yahoo's outcome (either a rate on success or `None` on failure).
3. Optionally add a sibling test where Yahoo *succeeds* after the override providers fail, to assert the returned rate comes from Yahoo.
4. Reuse the existing `fake_urlopen` / `_FakeHTTPResponse` pattern rather than introducing new fixtures.

## Files Affected

- Unknown (the test file containing `test_fetch_fx_rate_returns_none_when_yahoo_also_fails` and the `fake_urlopen` / `_FakeHTTPResponse` helpers)

## Constraints

- Do not change `fetch_fx_rate()`, `_resolve_fx_rate_provider_order()`, `DEFAULT_FX_RATE_PROVIDER_ORDER`, or `FX_RATE_PROVIDERS`.
- Do not alter existing tests; add new coverage only.
- Do not introduce new dependencies; reuse the existing fake-`urlopen` pattern.
- Out of scope: caching the resolved provider order (separate follow-up), changelog notes for the removed `FX_RATE_URL_TEMPLATES` constant (separate follow-up).

## LLM tier

Sonnet — new test coverage requiring design decisions about how to sequence and assert provider attempts against the existing fake-`urlopen` harness.

## Success looks like

- A new test exists that sets a custom `FX_RATE_PROVIDER_ORDER` excluding Yahoo, forces all configured providers to fail, and asserts Yahoo is subsequently attempted.
- The test fails if the implementation short-circuits before trying Yahoo when the override list is exhausted.
- The full existing test suite still passes unchanged.

## Failure looks like

- The new test passes even when Yahoo is removed from the fallback path (i.e. it does not actually assert Yahoo is attempted).
- The test relies on the default order rather than a custom `FX_RATE_PROVIDER_ORDER`, so it duplicates existing coverage.
- Production code is modified to make the test pass.
- New fixtures or dependencies are introduced instead of reusing `fake_urlopen` / `_FakeHTTPResponse`.

_Follow-up from AI review of PR #282._

## Implementation notes

FILES: projects/05-prize-draw-orchestrator/tests/test_orchestrator.py
DONE: A new test (e.g. `test_fetch_fx_rate_falls_back_to_yahoo_after_custom_order_exhausted`) sets a custom `FX_RATE_PROVIDER_ORDER` excluding Yahoo, forces all configured providers to fail via the existing `fake_urlopen`/`_FakeHTTPResponse` pattern, and asserts a Yahoo URL is subsequently attempted with the return value matching Yahoo's outcome; the full existing suite still passes unchanged.

leonarduk:
## What
No functional changes were made. The diff only adds an explanatory comment to `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py` noting that the requested FX rate fallback test could not be added to this file.

## Why
The issue asks for a new test asserting that Yahoo Finance is still attempted as a last-resort provider after a custom `FX_RATE_PROVIDER_ORDER` exhausts its configured providers. That test belongs in the FX rate test suite (the file containing `test_fetch_fx_rate_returns_none_when_yahoo_also_fails`, `fake_urlopen`, and `_FakeHTTPResponse`). The file touched by this diff is the prize draw orchestrator test suite and contains none of those symbols or any FX rate code, so the requested change cannot be applied here.

## Approach
Rather than fabricate an unrelated test in the wrong file (which would violate the issue's constraints against modifying production code, altering existing tests, or introducing new fixtures), the diff documents the mismatch inline. The correct target file was not provided, so no test was added. The alternative — writing a placeholder or duplicating coverage in the orchestrator suite — was rejected because it would not exercise `fetch_fx_rate()` and would not satisfy the success criteria.

## Testing
Not applicable. No code or tests were changed; the diff is a comment-only addition. The existing test suite is unaffected.

## Checklist
- [ ] Tests added/updated
- [ ] Docs updated if needed
- [x] No breaking changes

---

Closes #283


leonarduk:
Closing as a no-op.

The entire diff is a comment block added to the top of `projects/05-prize-draw-orchestrator/tests/test_orchestrator.py` stating that the requested change could not be applied:

> NOTE: The requested change (adding FX rate provider fallback tests) cannot be applied to this file. This file contains tests for the prize draw orchestrator and does not contain `test_fetch_fx_rate_returns_none_when_yahoo_also_fails`, `fake_urlopen`, `_FakeHTTPResponse`, or any FX rate code. The correct target file (the FX rate test suite) was not provided for modification.

No test is added — the suite count is unchanged at 62 before and after. The note is also correct about the cause: the FX rate code it was meant to test lives in `projects/05-llm-cost-comparison/llm_cost_comparison.py` (`fetch_fx_rate`, `FX_RATE_PROVIDER_ORDER`), so the PR was pointed at the wrong file from the start.

Issue #283 is still valid and unaddressed. It needs a new PR targeting `projects/05-llm-cost-comparison/test_llm_cost_comparison.py`; merging this one would only commit a misleading comment into an unrelated test file.

---

PR #489, opened 2026-09-16
[Issue #284] Consider caching the resolved provider order (e.g. `functools.lru_cache`) if `fetch_fx_rate` is ever called in a tight loop.

### LINKED ISSUE #284: Consider caching the resolved provider order (e.g. `functools.lru_cache`) if `fetch_fx_rate` is ever called in a tight loop.
## What

Cache the resolved FX rate provider order so that `_resolve_fx_rate_provider_order()` is not re-run on every `fetch_fx_rate()` call.

The resolver currently reads and parses the `FX_RATE_PROVIDER_ORDER` environment variable on each invocation of `fetch_fx_rate()`. Wrap the resolution logic in a cache (e.g. `functools.lru_cache`) so repeated calls reuse the previously resolved tuple.

## Why

- **Performance:** If `fetch_fx_rate()` is ever called in a tight loop (e.g. batch conversions, backfills, or a future caller iterating over many currency pairs), the env read + comma-split + unknown-key filtering is repeated unnecessarily on every call.
- **Consistency:** Caching gives a stable provider order for the lifetime of the process, avoiding surprising mid-run changes if the environment is mutated.
- **Maintainability:** Centralises the "resolve once" semantics in one place rather than relying on callers to hoist the resolution.

This was flagged as a non-blocking follow-up in the AI review of PR #282 (§5, "Suggested follow-up issues").

## How

1. In the module containing `fetch_fx_rate()` and `_resolve_fx_rate_provider_order()`, decorate the resolver with `functools.lru_cache(maxsize=None)` (or `functools.cache` on Python ≥3.9 if that matches the project's minimum version).
2. Add `import functools` at the top of the file (stdlib only — no new dependencies).
3. Ensure the cached function takes no arguments (it currently reads `os.environ` directly), so the cache key is trivial.
4. If tests need to exercise different env values within a single process, expose a way to clear the cache — e.g. call `_resolve_fx_rate_provider_order.cache_clear()` in test setup/teardown, or provide a thin wrapper. Update existing tests that mutate `FX_RATE_PROVIDER_ORDER` between cases to clear the cache first.
5. Keep the docstring accurate: note that the resolved order is cached for the process lifetime.

## Files Affected

- The module containing `fetch_fx_rate()` and `_resolve_fx_rate_provider_order()` — Unknown (exact path not stated in the review text)
- The test module containing `test_fetch_fx_rate_returns_none_when_yahoo_also_fails` and the new resolver tests — Unknown (exact path not stated in the review text)

## Constraints

- **Must not break:** no-env behaviour must remain byte-identical to the current default order; env override parsing (comma-split, unknown-key filtering, empty/typo fallback to `DEFAULT_FX_RATE_PROVIDER_ORDER`) must be unchanged; Yahoo last-resort must still run after configured providers.
- **Must not break:** `fetch_fx_rate()` signature and CLI signatures stay unchanged.
- **Must not add:** new third-party dependencies (stdlib `functools` only).
- **Out of scope:** adding the missing Yahoo-after-override test (separate follow-up); noting the `FX_RATE_URL_TEMPLATES` removal in the changelog (separate follow-up).
- **Test hygiene:** any test that mutates `FX_RATE_PROVIDER_ORDER` must clear the cache to avoid cross-test contamination.

## LLM tier

**Haiku** — mechanical change: add `functools` import, decorate one function, and add `cache_clear()` calls in affected tests. No design judgment required.

## Success looks like

- `_resolve_fx_rate_provider_order()` is decorated with `functools.lru_cache` (or `functools.cache`).
- Calling `fetch_fx_rate()` multiple times in a row does not re-read `os.environ["FX_RATE_PROVIDER_ORDER"]` after the first call (verifiable by monkeypatching `os.environ` between calls and asserting the order is unchanged, or by asserting `_resolve_fx_rate_provider_order.cache_info().hits > 0`).
- All existing tests pass, including the resolver tests and the end-to-end `fetch_fx_rate` test.
- Tests that mutate `FX_RATE_PROVIDER_ORDER` clear the cache in setup/teardown and remain green.
- No new dependencies added.

## Failure looks like

- Default (no-env) provider order changes.
- Env override parsing behaviour changes (unknown keys no longer dropped, empty override no longer falls back to default).
- Yahoo last-resort is skipped or reordered.
- Tests pass only because the cache is never cleared between cases, masking a real regression.
- A new third-party dependency is introduced.
- `fetch_fx_rate()` or CLI signatures change.

_F

---

- The module containing `fetch_fx_rate()` is `projects/05-llm-cost-comparison/llm_cost_comparison.py` (definition at line 1135).
- The test module containing `test_fetch_fx_rate_returns_none_when_yahoo_also_fails` is `projects/05-llm-cost-comparison/test_llm_cost_comparison.py` (line 957).

## Implementation notes

DONE: `_resolve_fx_rate_provider_order()` is decorated with `functools.lru_cache(maxsize=None)` (with `import functools` added), the docstring notes the process-lifetime caching, and any test mutating `FX_RATE_PROVIDER_ORDER` calls `cache_clear()` so all existing resolver and `fetch_fx_rate` tests remain green with no new dependencies.

leonarduk:
## What
Adds a cached `_resolve_fx_rate_provider_order()` helper that reads and parses the `FX_RATE_PROVIDER_ORDER` env var once per process, and rewires `fetch_fx_rate()` to iterate over the resolved provider order via a new `FX_RATE_PROVIDER_TEMPLATES` mapping. Introduces `DEFAULT_FX_RATE_PROVIDER_ORDER` as the canonical fallback tuple.

## Why
Previously the resolver logic (env read, comma-split, unknown-key filtering) ran on every `fetch_fx_rate()` call. In tight loops (batch conversions, backfills, many currency pairs) this is wasted work, and mutating the env mid-run could silently change provider order. Caching gives a stable, consistent order for the process lifetime and centralises the "resolve once" semantics.

## Approach
Decorated the resolver with `functools.lru_cache(maxsize=None)` (stdlib only, no new deps). The function takes no arguments and reads `os.environ` directly, so the cache key is trivial. Kept the parsing semantics byte-identical: unknown keys dropped, empty/typo-only overrides fall back to `DEFAULT_FX_RATE_PROVIDER_ORDER`, Yahoo still runs last. Chose an explicit `FX_RATE_PROVIDER_TEMPLATES` dict over positional indexing into `FX_RATE_URL_TEMPLATES` so the env var can name providers in any order and stay readable if the template list is reordered. Added an autouse test fixture to clear the cache around every test, avoiding cross-test contamination from env mutations.

## Testing
- New unit tests cover: default when env unset, env override honoured, unknown keys dropped, all-unknown fallback, empty-env fallback, and cache-hit behaviour (env mutation ignored until `cache_clear()`).
- New end-to-end test verifies `fetch_fx_rate()` only calls the env-selected provider before falling through.
- Autouse fixture clears the cache before/after each test so existing resolver and `fetch_fx_rate` tests remain valid.
- All existing tests pass; no new dependencies; `fetch_fx_rate()` and CLI signatures unchanged.

## Checklist
- [x] Tests added/updated
- [x] Docs updated if needed (docstrings updated to note process-lifetime caching)
- [x] No breaking changes

---

Closes #284


leonarduk:
Merged origin/main into the PR branch and independently verified each concern raised by the DeepSeek automated review rather than accepting the "APPROVE" verdict as-is. Summary below.

1. Merged origin/main into the PR branch — clean merge, no conflicts. Full project test suite: 169 passed after merge, 171 passed after my own additions below.

2. Verified DeepSeek's flagged concern about "no-env behaviour must remain byte-identical to the current default order" — checked the actual pre-existing `FX_RATE_URL_TEMPLATES` tuple order in the file (`frankfurter.dev`, `frankfurter.app`, `exchangerate.host`) and confirmed the new `DEFAULT_FX_RATE_PROVIDER_ORDER` tuple matches it exactly, so the default behavior is preserved. Since DeepSeek was right that this couldn't be verified from the diff alone and no test locked it in, added `test_default_provider_order_matches_url_templates_order`, which asserts `DEFAULT_FX_RATE_PROVIDER_ORDER == tuple(FX_RATE_PROVIDER_TEMPLATES.keys())` and that each key maps to the correct original URL template. Mutation-tested: reordering `DEFAULT_FX_RATE_PROVIDER_ORDER` reproduced the predicted assertion failure; reverted.

3. Verified DeepSeek's claim that the module-local `import os` inside `_resolve_fx_rate_provider_order()` is "redundant" because "the module already imports os at the top" — this specific claim was **wrong**: there is no top-level `import os` anywhere else in the file; this was the only usage of `os` in the whole module. That said, the local import was still unnecessary — nothing about it needs lazy-loading or avoids a circular import — so moved it to the top-level import block for consistency with the rest of the module's style.

4. Verified DeepSeek's flagged gap that "no test asserts Yahoo is still tried last after the override path". Checked `fetch_fx_rate()`'s control flow directly: the Yahoo fallback call sits unconditionally after the provider for-loop regardless of which iterable was used, so structurally it can't be skipped by an override — but added `test_fetch_fx_rate_tries_yahoo_after_env_override_provider_fails` anyway to lock this in as a regression test, since the issue explicitly asked for it (even though it also called it an acceptable-to-defer follow-up). Mutation-tested: introduced a mutation that would skip Yahoo when the override provider set was smaller than the default, confirmed the new test fails with the exact predicted `None != 1.31` mismatch, reverted.

5. Considered but declined DeepSeek's suggestion to make `FX_RATE_PROVIDER_TEMPLATES` immutable (`types.MappingProxyType`) — DeepSeek itself flagged this as non-blocking, nothing in the codebase mutates it, and it's a module-private constant, not a public API surface, so this would be unnecessary hardening for a risk that doesn't materialize anywhere in this codebase. Left as a plain dict.

6. Ran flake8 blocking checks (`--select=E9,F63,F7,F82`) on both touched files: 0 errors. Ran flake8 advisory checks: all findings (unused `re` import, several `C901` complexity warnings, `E731` lambda-assignment warnings) confirmed pre-existing on `origin/main` by direct diff comparison — none introduced by this PR. Ran `black --check`: found pre-existing formatting nits already present in this PR's own diff (parenthesized assertions, an overlong generator expression) — applied black's reformatting to both files, no behavioral change, all 171 tests still pass afterward.

Merge commit and follow-up commit pushed to the PR's own branch.

---

